PR #2, opened 2023-12-20
Add Copilot commands for inline chat and chat sidebar

pokey:
Contains some very experimental commands for interacting with Copilot, several of which integrate with Cursorless targets. See included README for more info


C-Loftus (approved):
Looks good, thank you for this Pokey.

PR #5, opened 2023-12-29
Adds a command that allows you to answer in line

jaresty:
- Sometimes you want to answer right below your question
- This allows you to do that, to replace the question, or to answer above

### DIFF
--- a/GPT/gpt.talon
+++ b/GPT/gpt.talon
@@ -1,12 +1,6 @@
 mode: command
 
 -
-model answer {user.makeshift_destination} <user.cursorless_target>:
-    text = user.cursorless_get_text(cursorless_target)
-    result = user.gpt_answer_question(text)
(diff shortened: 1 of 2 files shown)


C-Loftus:
I am not seeing where the makeshift destination is defined. Let's chat about this later. I like the idea of combining model operations directly alongside cursorless, since it helps with some things that Copilot might not do as well for plain text. As it looks now, I'm not sure if it is a bit too specific to be merged. It would be helpful for you to discuss a concrete example of how others could use this as well, since I don't think many people would have a question or prompt directly written in their code. However, I could be overlooking something since I don't use the Emacs workflow you described on Slack.

I recently made a change regarding the API, so you may need to briefly redo this. Sorry about that.

C-Loftus:
Looks good after our chat on Discord. Merged. Thanks for your work.

PR #6, opened 2023-12-29
Implemented proper cursorless command to insert

jaresty:
- This handles multiple inputs
- It can also be used to insert the response to an arbitrary location


C-Loftus:
Looks great, thanks for figuring that out

PR #8, opened 2024-01-27
Merge branch 'main' of https://github.com/C-Loftus/talon-ai-tools

Mark-Phillipson:


C-Loftus:
Since we already have this command could you change your commit from 

```
pilot previous: user.vscode("editor.action.inlineSuggest.showPrevious")
pilot last: user.vscode("editor.action.inlineSuggest.showPrevious")
```
to be 

```
pilot (previous | last) : user.vscode("editor.action.inlineSuggest.showPrevious")
```
So we have less duplication. Thanks!  I think the phrasing is fine though and certainly makes sense based on the action name

( I am assuming the fact that you saw this command was defined already but just want to change an additional phrasing)

Mark-Phillipson:
Surprisingly I did not notice the command was already there.

Changed as You Suggested...

C-Loftus:
Sounds good, merged.

PR #9, opened 2024-01-29
Don't smash shell lines together

pokey:


pokey:
> make a shell helper through the copilot interface

can you elaborate?

C-Loftus:
```
pilot shell <user.text>:

jump to copilot tab 
Paste some automatic prompt boiler plate that signifies it is related to the shell 
insert the rest user text as the rest of the prompt
```

so the user could say `pilot shell  print all the files in the current working directory`

 but maybe this isn't a good idea since even dictating shell commands is kind of difficult at a high level,  since it's difficult in one phrase to say everything you want

pokey:
Maybe add a prompt in the talon list, and as a safety mechanism strip new lines if we detect we're in a terminal?

C-Loftus:
Yeah so currently we do have something pretty similar,  it's just going to obviously be operating upon copied text instead of arbitrary prompts from the user given through the speech command. ( so is pretty safe since its more likely to be used in something like `.bashbook` but honestly is probably not that useful)
I think what I will do honestly is create a safety mechanism through an imgui popup that the user has to explicitly confirm,  since I'm a little uncomfortable matching to a terminal tag since I think there are quite a few to match on. 

https://github.com/C-Loftus/talon-ai-tools/blob/b2e320a3fefefd7bfe427f177ef8fa8f8d2db5e7/GPT/staticPrompt.talon-list#L24

C-Loftus:
Pushed some new functionality.  I am fairly happy with this,  it's good for simple one liners and we don't have to strip any \n or \t  or worry about context. 

Eventually I will probably integrate this with fzf or shell autocomplete in some capacity, but need to think about a way to do that privately if that is even possible

```
model shell <user.text>$:
    result = user.gpt_generate_shell(user.text)
    user.add_to_confirmation_gui(result)

# Confirm and paste the output of the model
^paste model output$:
    user.paste_model_confirmation_gui()

^copy model output$:
    user.copy_model_confirmation_gui()

# Deny the output of the model and discard it
^deny model output$:
    user.close_model_confirmation_gui()
```

pokey:
Yep just tested and works well! A bit clunky, but if we find ourselves using it a lot we can streamline. See minor comment too https://github.com/C-Loftus/talon-ai-tools/commit/62a1ad117ef5433462eb6dddb59bd6150683688f#r138164479

C-Loftus:
Thanks for catching that last comment. I forgot to change that when refactoring.

Yeah, I think for the time being this is probably sufficient for the shell. I will continue looking into some other options for how we may be able to improve UX or extend AI shell functionality with shell integration directly.

jaresty:
I think it's probably a good idea, given that we aren't working together directly, to either have a separate development branch or a pull request flow just to keep main stable for casual users.

pokey:
yeah I think let's go for pull request flow and if it becomes cumbersome we'll revisit. Between the 3 of us I think we can keep things moving

PR #13, opened 2024-02-05
use gpt functions to handle different needs

jaresty:
Adds function calling to the ChatGPT can interact with the user's system intentionally.


C-Loftus:
For pokey's context. We reviewed this in a call. I plan to look this over in the coming days and tweak a few things. I may change the function calling to be more explicit and reduce some duplication in the HTML help menu. 

Lots of potential with function calling generally though.

C-Loftus:
Other thing I forgot to mention. It seems we have some changes coming just from different python formatters (changes of " to ' or vice versa, spacing etc). I might set up a pre-commit yml to have a standardized autoformat so these aren't treated as changes. Let me know if you have a preference or if this is due to other reasons. (I have to look into how this is actually set up since I haven't done this logistically before, can't be too hard unless Im misunderstanding something)

jaresty:
Oh yeah I don't have strong opinions on formatting in python. I think I had run the formatter once because I had some trouble with pasting a properly indented chunk of Json.

C-Loftus:
Checked this out in more detail. Here is what I would prefer if you wouldn't mind refactoring:

- I think we should have a separate `gpt-fuction-calling.py` file and a separate public function just for function calling. The gpt.py file is getting a bit verbose and I would prefer not passing a function object. I think to start we should see how the function calling does, and slowly but surely expand things. I am worried about any implicit behavior as we expand this. 
    -  Instead of having `actions.user.search_for_command()` be global we can just scope it privately to that file since it doesn't need to be globally exposed and having a static function instead of an overloadable talon function also reduces error potential. 
- Instead of using `model ask`, I was thinking of having an explicit separate command for function calls like
```
model intuit <user.text>: 
        instruction = user.text
        user.gpt_function_call(instruction)
```
- This separate function will allow us to keep the talon file commands in the talon files like user.paste so it is more clear to the user what is occurring with each model command. Anything with keyboard input is important to be explicit in my opinion
- I don't think we need `display_response`. I believe we can just use gpt_help since it is the same code? If we need a natural language and cleaner function name for function calling, we can just call wrap `actions.user.gpt_help`

jaresty:
I am mostly fine with the suggestions above but i have two main areas I disagree:

- with respect to the function calling for input - the LLM does a much more targeted response when given a function to insert a response. If you compare the difference between "model ask insert a while loop" versus "model ask how do I do a while loop" I think you will see what I mean. I think the usability difference warrants the change.

- for the display command - it's not exactly the same code as got_help. The body of the resulting page is the contents of the LLM response. I think this is useful when the LLM is responding to a user and you don't want to paste the text into the window since it is more conversational. Are you suggesting that gpt_help be changed to take that text and display it instead?

jaresty:
The one issue with making a separate command for function calls is that it will require a different request be sent so there could be some duplication there. 

As to a name, how about "model do" since we're asking the model to do what we ask and letting it decide?

jaresty:
OK! Responded to your feedback. There is now a `model go` command that will execute function calls if they are requested.

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+from typing import Callable, Literal
+import webbrowser
+import tempfile
+import requests
+import os
+import json
+
+mod = Module()
+
+mod.setting(
```
C-Loftus, line 11: We don't need to duplicate this setting. Can just get rid of it in this file

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+    "openai_model", type=Literal["gpt-3.5-turbo", "gpt-4"], default="gpt-3.5-turbo"
+)
+
+# Defaults to Andreas's custom notifications if you have them installed
+def notify(message: str):
```
C-Loftus, line 16: Can just import this from gpt.py using 

`from .gpt import notify`

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+
+    response = requests.post(url, headers=headers, data=json.dumps(data))
+
+    if response.status_code == 200:
+        notify("GPT Task Completed")
+        print(response.json())
+
+        try:
```
C-Loftus, line 131: Perhaps refactor this into a separate function just for clarify

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+            tool_calls = response.json()["choices"][0]["message"]["tool_calls"]
+            while tool_calls:
+                tool = tool_calls.pop()
+                first_argument = tool['function']['arguments']
+                try:
+                    first_argument = json.loads(first_argument)['str']
+                except Exception as e:
+                    notify(f"Argument Jason was malformed: {e}")
+                if tool['function']['name'] == 'insert':
```
C-Loftus, line 140: I like `match` statements for this type of situation, but `if` works too

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+def notify(message: str):
+    try:
+        actions.user.notify(message)
+    except:
+        app.notify(message)
+    # Log in case notifications are disabled
+    print(message)
+
+
+def gpt_query(prompt: str, content: str, insert_response: Callable[[str], str]) -> str:
```
C-Loftus, line 25: even though this is in a separate namespace let's rename it, to `gpt_function_query` or something similar just so we don't get confused by clashing the function names.

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+            actions.user.display_response(content)
+
+    else:
+        notify("GPT Failure: Check API Key, Model, or Prompt")
+        print(response.json())
+
+
+@mod.action_class
+class UserActions:
+    def gpt_go(utterance: str, selected_text: str) -> str:
```
C-Loftus, line 162: I would rename this to `gpt_call_function` or something similar that you like that more clearly specifies what it does. I think `go` is a bit unclear to new users.

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,229 @@
...
+                "content": f"language:\n{actions.code.language()}",
+            },
+            {"role": "user", "content": f"{prompt}:\n{content}"},
+        ],
+        "tools": [
+            {
+                "type": "function",
+                "function": {
+                    "name": "insert",
+                    "description": "insert(str: string) - this inserts the string into the document. The document is in the language specified so if you aren't careful you will case syntax errors.",
```
C-Loftus, line 52: Small typo, 'case' instead of 'cause'

C-Loftus:
Thanks for your work.  I added some comments.  relatively simple refactoring stuff.

 I'm thinking as well that we might want to rename `gpt go` to `gpt call` since that is also a one syllable verb and it makes a little bit more sense semantically.  So a user could say `model call a function that opens the file panel` for vscode.  But you have more familiarity with how the prompts are constructed so maybe `go` makes more sense with how the wording would be passed in

C-Loftus:
Essentially my thought is that until we have high confidence in the fact this function calling is working well,  we want to be as explicit as possible. I was presuming but am not entirely sure if we will have better performance if the user crafts the text in such a way that they are aware of the underlying function calling.

 The only real issue I have with `model go` is that it appears a bit semantically ambiguous and not entirely clear how it differs from the `please` command.

You definitely have more experience than I do within function calling so I will let you make the final decision,  but that was my rationale

jaresty:
I've been using it like this:

- "model go insert a while loop"
- "model go please split this window"

Swapping in call is a bit awkward:

- "model call insert a whole loop"
- "model call please split this window"

I think the similarity to please is a feature; the function calling is an implementation detail that I don't really want to expose since you are using it to solve a task (not for the purpose of calling a function)

C-Loftus:
Ok let's go with that. And if we have any other ideas we can obviously edit before merge. I will probably ask for Pokey's feedback before merge.

GPT/gpt-function-calling.talon
```diff
@@ -0,0 +1,15 @@
+mode: command
+tag: user.cursorless
```
C-Loftus, line 2: This scopes all the go commands to just cursorless. Is that what you want?  I wasn't sure if you want to only function call within visual studio code at the moment.

jaresty: One of the commands requires cursorless, so I wasn't sure what to do there.  I didn't really want to split out a separate file. What do you think?

jaresty:
I updated this to use the new builder and squashed the commits.

GPT/gpt-function-calling.talon
pokey: I would split it. No reason the other command shouldn't be available outside of cursorless. We could have one `.talon` file where we put all the cursorless stuff if you wanna avoid lots of small files. For that matter, I'm not sure the other command in this file needs its own file

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,185 @@
...
+    def notify_user(response: str):
+        """Send a notification to the desktop"""
+        actions.app.notify(response)
+
+    def search_for_command(response: str):
+        """Search VSCode for command"""
+        actions.user.command_palette()
+        actions.user.paste(response)
+
+    def display_response(response: str):
```
pokey, line 173: would be cool to let gpt just output arbitrary html to be rendered. need to think about security implications and def out of scope for this PR

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,185 @@
...
+        notify(f"No tool_calls found in response from LLM: {e}")
+
+@mod.action_class
+class UserActions:
+    def gpt_go(utterance: str, selected_text: str) -> str:
+        """Answer an arbitrary question"""
+
+        return gpt_function_query(utterance, selected_text, actions.user.paste)
+
+    def gpt_go_cursorless(utterance: str, text_to_process: str, cursorless_destination: any):
```
pokey, line 158: you're using `cursorless_get_text_list` so `text_to_process` has type `list[str]`. This makes sense if you want to support multiple targets (eg `"model go <prompt> air and bat and cap after drum"` or `"model go <prompt> every token after bat"`.

If you just want to support single target, you can use `cursorless_get_text` instead

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,185 @@
...
+                    actions.user.notify_user(first_argument)
+                case 'search_for_command':
+                    actions.user.search_for_command(first_argument)
+    except Exception as e:
+        notify(f"No tool_calls found in response from LLM: {e}")
+
+@mod.action_class
+class UserActions:
+    def gpt_go(utterance: str, selected_text: str) -> str:
+        """Answer an arbitrary question"""
```
pokey, line 154: I have mixed feelings about "go" as a verb, but I'm not strongly opposed. If we do go with such a vague verb, then we really need a good docstring here

GPT/gpt-function-calling.talon
jaresty: 👍 Sounds good to me. I pushed these changes.

GPT/gpt-function-calling.py
jaresty: Updated the doc string

jaresty: I'm not so sure about the choice of the word go myself, but I don't see this as intentionally different from `model please` for example. The user is asking the model to do something on their behalf, and they don't really care how it is done. The function calling is an implementation detail that simply makes the LLM more aware of the impact it's having on the user's system. For that reason, the LLM can more intentionally switch between inserting text, displaying text, running commands, and sending notifications. I like having something that is ambiguous and quick since the function calling is not the point.

GPT/gpt-function-calling.py
```diff
@@ -0,0 +1,185 @@
...
+
+    response = requests.post(url, headers=headers, data=json.dumps(data))
+
+    if response.status_code == 200:
+        notify("GPT Task Completed")
+        print(response.json())
+
+        process_function_calls(insert_response, response)
+
+        content = (response.json()["choices"][0]["message"]["content"] or "").strip()
```
pokey, line 121: Let's store `(response.json()["choices"][0]["message"]` in a variable so we don't have to duplicate it below?

GPT/gpt-function-calling.py
pokey: so I guess it just gets turned into a string representation of the list I guess. I'd make the type here be `list[str]` and then explicitly convert it to a list here.

GPT/gpt-cursorless.talon
```diff
@@ -9,4 +9,7 @@ tag: user.cursorless
     result = user.gpt_apply_prompt(user.staticPrompt, text)
     user.cursorless_insert(cursorless_destination, result)
 
-
+# Apply a prompt to any text, and output it any target
+^model go {user.staticPrompt} <user.cursorless_target> <user.cursorless_destination>$:
```
pokey, line 13: do we want the same set of prompts here or do we want prompts specific to `go`?

GPT/gpt.talon
```diff
@@ -7,6 +7,12 @@ model ask <user.text>$:
     result = user.gpt_answer_question(text)
     user.paste(result)
 
+# Say your prompt directly and the AI will apply it to the selected text
+model go <user.text>$:
```
pokey, line 11: do we want to support static prompts as well?

GPT/gpt-cursorless.talon
jaresty: I don't think there's a need to keep the prompts separate. This goes back to the point that we are not intending to do function calling here; we're simply allowing the user to ask anything and let the LLM make a call as to what to do in response.

GPT/gpt.talon
jaresty: This is the primary way that I use this, I do want to allow this.

jaresty: I misunderstood at first. Maybe? Good question.

C-Loftus: Not sure if it makes sense to have static prompts unless you want to make function calling less explicit. Since if a user wants to apply a static prompt they can just say the name.

GPT/gpt-cursorless.talon
pokey: Seems fine for now. I also wonder if we want target and/or destination to be optional

pokey: But I guess all of these prompts expect input text? So I guess makes sense to require both

GPT/gpt.talon
jaresty: I think it does make sense to have static prompts because the function calling is not actually different except in the level of trust you are giving to the LLM.  This is why I originally didn't have this as a separate Python file-there isn't a different use case. It's an implementation detail.

GPT/gpt-cursorless.talon
jaresty: I could see an argument to make destination optional. In that case it would just paste the response instead of putting it into the destination in case of an insertion function call.

jaresty: Could you give me an example of the Python code to handle the optional destination there? I'm not sure exactly how it would appear to Python if the user doesn't supply it.

GPT/gpt-function-calling.py
C-Loftus: Just as a thought, we obviously don't need to use model before `go`. I personally do not really like `go`. Something like `hey talon` may be interesting since it mimics the behavior of `hey siri` and more clearly signifies that there is some background guesswork

jaresty: Fwiw - I don't mind putting this behind the beta tag, if that is preferable.

jaresty: I asked ChatGPT for some suggestions and this is what it came up with:

🤖
Command Initiation 'Begin', 'Start', 'Commence'
🙏
Politeness 'Please', 'Kindly', 'Would you mind'
🗣️
Request 'Can you', 'Could you', 'Would you be able to'
🔍
Search 'Find', 'Search for', 'Look up'
📝
Writing 'Write', 'Compose', 'Draft'
📅
Scheduling 'Schedule', 'Set up', 'Arrange'
📧
Email 'Send an email', 'Compose an email', 'Draft an email'
📊
Analysis 'Analyze', 'Assess', 'Evaluate'
📌
Specificity 'Specifically', 'In particular', 'Regarding'

jaresty: I kind of like `model can you`; it's longer than `go` but pretty easy to say.

C-Loftus: `model can you` sounds good with me. It does start to sound similar to `model please`  and I still do like something like `hey $X`, but I don't want to be fastidious, so this is fine. 
We will just put it behind the beta tag until we align with user feedback

C-Loftus:
Got a git issue figured out so I can commit to this. (messing with pre-commit if you see errors)

Generally as I have trying this out, I don't think that someone who is new to using the repo will understand what this command can do and have it be practical for them without actually reading the source, since it is very dependent on the underlying function call options. We don't want users to have to experiment to see what will work.  I think function calling is a cool idea and I am all for iterating and starting small but I think the amount of iteration this would take to be practical would be a lot. (ie we need to prevent this from causing errors, have a more scalable way of generating json function specs, have a plan for how to make a large amount of potential function calls maintainable if the underlying talon fn changes in args, etc.)

@jaresty I am curious, have you found this to be practically useful so far? For me, the code generation via `pilot make <phrase>` is much better since it has repo context and we don't need to pass in anything / prompt engineer on our end. 

Pokey, any opinions on this? As mentioned, I really appreciate the work Joshua has done on this, but I just am not sure if I see function calling as the best fit for us. Although perhaps I am overlooking something.

I just don't want to make sure that anything behind the beta tag has a plan for ways it could be feasibly stabilized and used practically.

jaresty:
I have found this to be useful. I think it can be expanded on as well. I don't have access to copilot presently so I am leaning on ChatGPT for all of my AI needs, and letting the AI differentiate between the kinds of things it can do is more functional for me than simply having it dump a response into the buffer directly. Also, function calling was improved in got-4 so I think OpenAI is leaning into this as one of their main directions going forward so I think it's good for us to experiment with it as well. As I mentioned before, I see a strong argument to make this the general use case instead of a separate entry point. I'm pretty sure that copilot is also built on top of function calling under the hood-so we ought to leverage the same IMO.

jaresty:
I'm trying to not get too attached to it until it is merged, however my expectation is that I will use this for almost all of my model usage. I find it useful to be able to use chatGPT to understand code and to insert syntax when I do not have a snippet and am not aware how to do it in the language I'm working in. I've been using this in python a lot. The experience of using chatGPT when it can display a response in a web browser is nicer than when it pastes into my buffer. I do think there is a world in which we might want to split out what functions are available for what sub commands; perhaps running a VSCode command should be its own thing. Because my experience is not the general case, I don't think we'll really know until more people try this out.

C-Loftus:
For context: aligned on a call. Seems to be in a good spot now, we will just scope the command under the beta tag for the time being. I think there is a path forward in terms of maintainability as well by splitting the functions into their own file and automatically generating the JSON.

GPT/gpt-cursorless.talon
jaresty: Thinking on this a bit more, I imagine that some prompts will only makes sense with function calling. Eventually we may want to split out a separate set of prompts for those. I'd like to support both the existing ones and some additional ones if we get there, but this is probably a future concern.

pokey: > Thinking on this a bit more, I imagine that some prompts will only makes sense with function calling. Eventually we may want to split out a separate set of prompts for those. I'd like to support both the existing ones and some additional ones if we get there, but this is probably a future concern.

yeah i think we can prob address that later if we come up with examples



> Could you give me an example of the Python code to handle the optional destination there? I'm not sure exactly how it would appear to Python if the user doesn't supply it.

You could either split it into two different commands, or wrap the destination in `[]` and then pass `cursorless_destionation or 0` as the arg to the action

GPT/gpt-function-calling.py
pokey: I actually really like the command "model please" for this one. Tbh I would be tempted to just have this new function-calling command replace the existing "model please" if you have the tag activated

Alternatives:

- `"model act"`
- `"model do"`
- `"model fly"`
- `"model auto"`
- `"model run"`

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -9,3 +9,8 @@ app: vscode
     txt = user.cursorless_get_text_list(cursorless_target)
     result = user.gpt_apply_prompt(prompt, txt)
     user.cursorless_insert(cursorless_destination, result)
+
+# Apply a prompt to any text, and output it any target
+^model can you {user.staticPrompt} <user.cursorless_target> <user.cursorless_destination>$:
+    text = user.cursorless_get_text_list(cursorless_target)
+    user.gpt_can_you_cursorless(user.staticPrompt, text, cursorless_destination)
```
pokey, line 16:
```suggestion
    text_list = user.cursorless_get_text_list(cursorless_target)
    user.gpt_can_you_cursorless(user.staticPrompt, text_list, cursorless_destination)
```

GPT/beta-commands/beta-func-calling.talon
pokey, line 1: I'm not sure the comments in this file are accurate

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+        "temperature": 0.6,
+        "n": 1,
+        "stop": None,
+        "model": settings.get("user.openai_model"),
+    }
+
+
+    response = requests.post(url, headers=headers, data=json.dumps(data))
+
+    message = response.json()["choices"][0]["message"]
```
pokey, line 113: shouldn't we wait to do this until we've established the status code is 200?

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+        print(response.json())
+        process_function_calls(insert_response, message)
+
+        content = (message["content"] or "").strip()
+        if len(content) != 0:
+            display_response(content)
+
+    else:
+        notify("GPT Failure: Check API Key, Model, or Prompt")
+        print(response.json())
```
pokey, line 125: we should prob throw an error here to stop execution of further talon commands if we support chaining in the future / they remove anchors

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+
+def process_function_calls(insert_response, message):
+    try:
+        tool_calls = message["tool_calls"]
+        while tool_calls:
+            tool = tool_calls.pop()
```
pokey, line 131: sorry i'm nitpicking here, but rather than mutating the argument, it's more idiomatic to iterate the list: `for tool_call in tool_calls:`

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+            first_argument = tool['function']['arguments']
+            try:
+                first_argument = json.loads(first_argument)['str']
+            except Exception as e:
+                notify(f"Argument Jason was malformed: {e}")
```
pokey, line 136:
```suggestion
                notify(f"Argument json was malformed: {e}")
```

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+                case 'insert':
+                    insert_response(first_argument)
+                case 'display':
+                    display_response(first_argument)
+                case 'notify':
+                    notify_user(first_argument)
+                case 'search_for_command':
+                    search_for_command(first_argument)
+    except Exception as e:
+        notify(f"No tool_calls found in response from LLM: {e}")
```
pokey, line 147: we prob want this exception to wrap just the statement where we extract tool_calls and `return` from here

GPT/beta-commands/gpt-function-calling.py
pokey, line 136: Do we really want to swallow the error? I would think we'd want to re-throw, or just not bother catching the error at all and letting it bubble out

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,161 @@
...
+
+@mod.action_class
+class UserActions:
+    def gpt_can_you(utterance: str, selected_text: str) -> str:
+        """Run a query against ChatGPT and allow it to execute targeted function calls on your machine"""
+        return gpt_function_query(utterance, selected_text, actions.user.paste)
+
+    def gpt_can_you_cursorless(utterance: str, text_to_process: str, cursorless_destination: any):
+        """Apply a cursorless prompt"""
```
pokey, line 156: not sure this docstring is accurate

GPT/beta-commands/gpt_callables.py
```diff
@@ -0,0 +1,45 @@
+from talon import actions
+from ..lib.HTMLbuilder import Builder
+from typing import Callable
+import inspect
+
+def _generate_tool(func: Callable):
```
pokey, line 6: this appears to be unused?

pokey (commented):
Ok left a few more comments

As for overall direction, I'm super excited about this approach. I think function calling is a really powerful way to combine the powers of Talon and GPT

GPT/beta-commands/gpt_callables.py
C-Loftus: To be used in the future. Right now, tool json specs are manually created, but we can automatically make them. I just don't have time to replace the generation code right now. I think leaving it in is fine for time being.

GPT/beta-commands/gpt-function-calling.py
C-Loftus: Changed. I think you're right the only thing I don't like about throwing an error is the fact we get a stack trace that we don't really need.

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -122,38 +121,44 @@ def gpt_function_query(prompt: str, content: str, insert_response: Callable[[str
...
+
+    for tool in tool_calls:
+
+        try: 
+            first_argument = tool['function']['arguments']
+            first_argument = json.loads(first_argument)['str']
+        except Exception as e:
+            notify(f"Argument JSON was malformed: {e}")
+            # Try next tool call if this one fails
+            continue
```
pokey, line 141: Is this desirable? If the first call fails, I'm not sure we want to blindly go on to the next one. Presumably the model put them in this order for a reason?

GPT/beta-commands/gpt_callables.py
pokey: Sounds good. Would be good to have a quick comment to that effect in the code

pokey (commented):
Almost there! See also older unresolved comments on this PR

pokey:
Strangely, I can't mark my own comments as resolved 🤔

GPT/beta-commands/gpt-function-calling.py
pokey: I don't think a stack trace hurts. Might be helpful for user to know where error is happening

GPT/gpt-cursorless.talon
jaresty: Thanks for that-I was able to make it optional. I think that makes it a lot more powerful.

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,165 @@
...
+                search_for_command(first_argument)
+
+
+@mod.action_class
+class UserActions:
+    def gpt_can_you(utterance: str, selected_text: str) -> str:
+        """Run a query with function calls and insert the result"""
+        return gpt_function_query(utterance, selected_text, actions.user.paste)
+
+    def gpt_can_you_cursorless(utterance: str, text_to_process: str, cursorless_destination: any):
```
pokey, line 159: note that `cursorless_destination` might be `0` now so we need to check for that case

GPT/beta-commands/gpt_callables.py
```diff
@@ -0,0 +1,118 @@
...
+                        },
+                        "required": ["str"],
+                    },
+                },
+            },
+            {
+                "type": "function",
+                "function": {
+                    "name": "display",
+                    "description": "DEFAULT - Display the response to the user. Use this for all informational text aside from notifications. Use this instead of returning content in the response.",
```
C-Loftus, line 88: Is there a reason we specify DEFAULT? Is it just a heuristic? I can't find this or anything similar in the openai docs

jaresty: I was trying to get the prompt to trigger the display without falling back to displaying the content manually. I think we could remove it if you like as it's not clear to me that it worked. I copied the example from another OSS project.

C-Loftus: Ok I think we can remove it

jaresty:
@C-Loftus  is there anything left to do on this one?

C-Loftus:
Other than resolving the merge conflict, I think it is pretty much fine. It would probably be good to have some sort of basic tests for this repo as we do the json schematization, but that is probably out of scope for this PR. 

I think we are just waiting for any feedback from Pokey. ( It is possible that this PR review system might be adding unnecessary friction, and if so I can do this merge without extra review.   Currently the repository is setup such that it requires a review to merge. But whatever is best for everyone )

jaresty:
Ah, I think you can review but not me-authors can't review their own PRs.

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,101 @@
...
+    response = requests.post(url, headers=headers, data=json.dumps(data))
+
+    if response.status_code == 200:
+        notify("GPT Task Completed")
+        message = response.json()["choices"][0]["message"]
+        process_function_calls(insert_response, message)
+
+        content = (message["content"] or "").strip()
+        if len(content) != 0:
+            display_response(content)
```
C-Loftus, line 48: We don't have to change this, but I personally feel like displaying the response in HTML isn't particularly useful unless its a help menu or something.

GPT/beta-commands/gpt-function-calling.py
```diff
@@ -0,0 +1,101 @@
...
+                "content": f"language:\n{actions.code.language()}",
+            },
+            {"role": "user", "content": f"{prompt}:\n{content}"},
+        ],
+        "tools": function_specs,
+        "max_tokens": 2024,
+        "temperature": 0.6,
+        "n": 1,
+        "stop": None,
+        "model": settings.get("user.openai_model"),
```
C-Loftus, line 36: Eventually will want to extract this to talon settings. Doesn't necessarily have to be in this PR though.

GPT/gpt-settings.talon
```diff
@@ -8,4 +8,4 @@ settings():
     user.openai_model = 'gpt-3.5-turbo'
 
 # Only uncomment the line below if you want experimental behavior to parse Talon files
-tag(): user.gpt_beta
+# tag(): user.gpt_beta
```
C-Loftus, line 11: Let's resolve this merge conflict so we use the `.example` config pattern instead of this.

GPT/beta-commands/gpt-function-calling.py
C-Loftus: Still need to fix this. Not exactly sure what we want to happen if the destination is invalid. I am not super clear on the usage.

C-Loftus (changes requested):
Just have to fix the merge conflict and the case where the destination is 0, then we should be fine to merge. If you want to add settings for temperature that is great too but otherwise we will do it later.

GPT/gpt-settings.talon
jaresty: done!

GPT/beta-commands/gpt-function-calling.py
jaresty: I have found this useful because it gives me feedback on my prompt.  I don't like to see the response pasted into the text buffer, but if I didn't succeed it getting a function to be called then I do like to see the response still.

C-Loftus: Ok we will leave it. I was thinking the imgui used for the shell might be better, but that isn't accessible so I am fine with whatever.

GPT/beta-commands/gpt-function-calling.py
jaresty: Handled the 0 case.

C-Loftus (dismissed):

jaresty:
@pokey it seems that the pre-commit.ci configuration and isort configuration are in conflict. No matter which one you conform to, the gpt-function-calling.py file doesn't pass CI.

pokey (commented):
Looking good! See comments below, and also see my pushed changes in 50aed65cbc50284346e3e12f05e845505bb963a0. And it looks like https://github.com/C-Loftus/talon-ai-tools/pull/13#discussion_r1484314812 got lost in all the comments

.pre-commit-config.yaml
```diff
@@ -30,10 +30,10 @@ repos:
...
     hooks:
       - id: flynt
-  - repo: https://github.com/pycqa/isort
-    rev: 5.13.2
-    hooks:
-      - id: isort
+  # - repo: https://github.com/pycqa/isort
+  #   rev: 5.13.2
+  #   hooks:
+  #     - id: isort
```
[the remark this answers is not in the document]
pokey: yep that fixed it

C-Loftus:
> Looking good! See my pushed changes in [50aed65](https://github.com/C-Loftus/talon-ai-tools/commit/50aed65cbc50284346e3e12f05e845505bb963a0). Also, it looks like [#13 (comment)](https://github.com/C-Loftus/talon-ai-tools/pull/13#discussion_r1484314812) got lost in all the comments

Awesome thanks for checking on this Pokey. Much appreciated. I personally sort of prefer `model can you` in reference to https://github.com/C-Loftus/talon-ai-tools/pull/13#discussion_r1484314812 since I still find `please` quite useful and prefer to have it be more deterministically still, even as we experiment with function calling.

But if you feel strongly otherwise I am fine changing it to `please`. Should be an easy change either way

GPT/gpt-function-calling.py
jaresty: I'm also a fan of "please"; is there an option that could allow @C-Loftus to keep the existing behavior if he wants it?

jaresty: Maybe the play is to make a separate tag for function calling to replace "please"?

C-Loftus:
I made some fairly significant changes to this. Previously, it pretty much never inserted text given the fact that insertion wasn't being passed to the API for function calling and as a result was only a backup. I turned insertion into another callable function and then made it so the insertion option is just passed as an enum.

For some reason previously we also had it pop up a new window every time.  I do not think that this is proper behavior so I changed it. If you want to have a pop up window in a new tab I think you should specify that in the prompt since otherwise it can get kind of annoying. 

With this functionality, it is now pretty much a superset of please so I'm okay with using the word please now and have changed it as such.

C-Loftus:
Cursorless functionality works without a destination, but only issue is that it seems like 
```
^model please <user.text> <user.cursorless_target> [<user.cursorless_destination>]$:
    text_list = user.cursorless_get_text_list(cursorless_target)
    user.gpt_dynamic_request_cursorless(user.text, text_list, cursorless_destination or 0)
```
doesn't work when a destination is defined. I get a weird ipc error that might be a signifier or a type issue outside this repo.

2024-02-17 14:38:41.562 ERROR    25:                                talon/scripting/talon_script.py:721| 
   24:                                talon/scripting/talon_script.py:314| 
   23:                                     talon/scripting/actions.py:88 | 
   22:  user/talon-ai-tools/GPT/beta-commands/gpt-function-calling.py:121| gpt_function_query(utterance, str(sele..
   21:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   20:  user/talon-ai-tools/GPT/beta-commands/gpt-function-calling.py:60 | process_function_calls(result, insert_..
   19:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   18:  user/talon-ai-tools/GPT/beta-commands/gpt-function-calling.py:97 | actions.user.cursorless_insert(first_a..
   17:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   16:                                     talon/scripting/actions.py:88 | 
   15:                   user/cursorless-talon/src/actions/actions.py:128| cursorless_replace_action(destination,..
   14:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   13:                   user/cursorless-talon/src/actions/replace.py:10 | actions.user.private_cursorless_comman..
   12:                                                                     ^
   11:                                     talon/scripting/actions.py:88 | 
   10:                           user/cursorless-talon/src/command.py:33 | actions.user.private_cursorless_run_rp..
    9:                                                                     ^
    8:                                     talon/scripting/actions.py:88 | 
    7:         user/cursorless-talon/src/cursorless_command_server.py:15 | actions.user.run_rpc_command_and_wait(..
    6:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    5:                                     talon/scripting/actions.py:88 | 
    4: user/knausj_talon/apps/vscode/command_client/command_client.py:307| run_command(
    3:                                                                     ^
    2: user/knausj_talon/apps/vscode/command_client/command_client.py:193| raise Exception(decoded_contents["error"])
    1:                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Exception: Cannot read properties of undefined (reading 'flatMap')
```

GPT/gpt-function-calling.py
C-Loftus: Changed to `please` since I changed `can you` to be a superset of please.

C-Loftus (approved):
Think this should be all set. I changed 
`str(selected_text)` to `" ".join(selected_text)` since casting an list of strings to a string  won't actually concat them together.  

and the cursorless flatmap type error was just due to an argument mix up. 

I'm fairly happy with this and I think it is ready to merge

jaresty:
Great! I'll kick the tires and then click merge when I'm back at my desk.

jaresty:
Quick question-it looks like insert response will always paste now; I wanted it to insert to the cursorless destination if it was provided. Am I misreading that?

jaresty:
I see it now-the switch is happening before the function calling.

jaresty:
Fwiw - the reason I had it defaulted to the pop up was that I prefer the textual responses from the LLM to display (versus editing). I think I'm ok with either for now since I'd like to get this merged sooner. We can always iterate on making behavior smoother later.

PR #14, opened 2024-02-06
GPT help menu

C-Loftus:
This commit improves the help menu to use a special builder class so reduce code reuse.

It also allows us to intelligently search through all the commands in the active context. It works okay, but there are some hallucinations since it doesn't understand the command syntax with regular expressions or Talon captures. That being said, for simple commands that directly match a word or phrase, it generally works. The downside is that we have to pass a ton of context in and it can take up to fifteen seconds in my experience.  I imagine this will also cost more for the user. I parallel process multiple requests since that is the only way to do all the commands at once, given the fact there is so much text to search through.

I am not sure if this should be merged since it isn't really reliable, however, it can be used as a heuristic to speed up learning Talon.

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -1,6 +1,8 @@
-from talon import Module, actions, clip, app, settings, imgui
+from talon import Module, actions, clip, app, settings, imgui, registry
 from typing import Literal
-import webbrowser, tempfile, requests, os, json
+import requests, os, json
+from .lib import HTMLbuilder
+from concurrent.futures import ThreadPoolExecutor
 
 mod = Module() 
 # Stores all our prompts that don't require arguments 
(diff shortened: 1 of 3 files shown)


jaresty:
Wow. I'm excited to give this a try.

jaresty:
One way to reduce the per-question cost could be to use a thread so you only have to upload the context when it changes. There is the question of when to initiate the upload.

C-Loftus:
Don't you have to always upload context for every request? How would it know what to search through? There isn't anything cached on their end to my understanding. And I think the active context could update very frequently as well.

Honestly this doesn't really work that well to begin with, but I didn't try GPT-4 since I don't have access. If you have that, I would say that might have much more promise

jaresty:
You should have ages to got 4 if you're using an API key. The threads functionality I mentioned is apparently called "assistants": https://platform.openai.com/docs/assistants/how-it-works/creating-assistants

C-Loftus:
I don't think so. I get the output 

```
2024-02-06 12:11:00.590    IO {'error': {'message': 'The model `gpt-4` does not exist or you do not have access to it. Learn more: https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}}   
```

I am not spending enough on this so I don't the majority of users will have access to this. I may just eliminate this for the time being. It was interesting to experiment with but I don't think that performance is good enough in general anyways. 

I think that honestly contributing more to traditional natural language documentation is probably more efficient and helpful for more people

jaresty:
I don't think that's the right name for the model. Here is a link to a zsh function that I've seen work before via the API: https://github.com/Aaronik/ai-functions/blob/76c51faebeece7974b148912f872ebf7dfec6147/ai.zsh#L42-L43

C-Loftus:
You can specify either the baseline name of the version with the preview. It is the proper name I'm fairly certain. The list of valid models are here:

https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo

jaresty:
I see. Can you give me an example how to use this? I'm happy to screen share. I tried it and the GPT request ended really fast but I didn't see any results.

jaresty:
I tried `model find split a window`

jaresty:
It looks like it's sending object names along with the command strings, was that your intention?

```
2024-02-06 10:45:28.687    IO {'messages': [{'role': 'user', 'content': '\n        The following is a list of commands for a program that controls the user\'s desktop. Each command is a series of words separated by a \'%\' character which is the separator to indicate the next separate command. I am a user and I want to find split a window. Return nothing but the specific command phrase that most closely matches this request. Do not return any other text or groups of unrelated words besides exactly the matching command. Each command is totally separate and unrelated to any word after a %. Special characters can be ignored. If there is no command return the exact word "None". \n        :\nterminal next"))%CommandImpl(Rule("terminal last"))%CommandImpl(Rule("terminal split"))%CommandImpl(Rule("terminal zoom"))%CommandImpl(Rule("terminal trash"))%CommandImpl(Rule("terminal toggle"))%CommandImpl(Rule("terminal scroll up"))%CommandImpl(Rule("terminal scroll down"))%CommandImpl(Rule("terminal <number_small>"))%CommandImpl(Rule("copy line down"))%CommandImpl(Rule("copy line up"))%CommandImpl(Rule("select less"))%CommandImpl(Rule("select (more | this)"))%CommandImpl(Rule("minimap"))%CommandImpl(Rule("maximize"))%CommandImpl(Rule("restore"))%CommandImpl(Rule("select breadcrumb"))%CommandImpl(Rule("replace here"))%CommandImpl(Rule("hover show"))%CommandImpl(Rule("join lines"))%CommandImpl(Rule("full screen"))%CommandImpl(Rule("curse undo"))%CommandImpl(Rule("curse red'}], 'max_tokens': 2024, 'temperature': 0.6, 'n': 1, 'stop': None, 'model': 'gpt-3.5-turbo'}
```

GPT/gpt.py
```diff
@@ -163,42 +165,53 @@ def gpt_help():
...
+                builder.p(result)
+        builder.render()
+
+
+def remove_wrapper(text: str):
+    wrapper = "CommandImpl('"
+    end_wrapper = "')"
+    if text.startswith(wrapper) and text.endswith(end_wrapper):
+        return text[len(wrapper):-len(end_wrapper)]
+    return text
```
C-Loftus, line 214: It should be removing that. 

In general, I am just going to close this since registry is unstable and this isn't really working well. Oh well, worth the experiment though

jaresty: That doesn't seem to be doing it. I think this is still worth experimenting with. I'd like to try to build on it if that's OK.

GPT/gpt.py
```diff
@@ -163,42 +165,53 @@ def gpt_help():
...
+        builder.render() 
+    
+
+    def gpt_find_talon_commands(command_description:str):
+        """Search for relevant talon commands"""
+        command_list = ""
+        for ctx in registry.active_contexts():
+            items = ctx.commands.items()
+            for _, command in items:
+                raw_command = remove_wrapper(str(command))
```
C-Loftus, line 185: That should be removing the string wrapper

GPT/gpt.py
C-Loftus: Ok feel free if you'd like. I personally won't be working on this since I am a bit frustrated with cross platform registry behavior

GPT/gpt.py
jaresty: It should be, but it did not.  I printed the request being sent and it didn't.  I was able to strip it with a regex:

```
def remove_wrapper(text: str):
    regex = r'[^"]+"([^"]+)"'
    match = re.search(regex, text)
    return match.group(1)
```

C-Loftus:
Talking on next PR

Issue #17, opened 2024-02-09
Add autoformatter


pokey:
we should be able to just steal the pre-commit setup from community

_Originally posted by @pokey in https://github.com/C-Loftus/talon-ai-tools/issues/13#issuecomment-1932274685_

C-Loftus:
I tried to set that up, but could not built talonfmt locally on windows due to treesitter. Must be something on my end, but didn't have success

pokey:
Interesting. In theory that shouldn't matter as it just needs to run in CI

Alternatively, we could start with just the Python stuff. That's the most important piece imo

PR #20, opened 2024-02-11
Add precommit config

C-Loftus:
Added the same settings as community. Not entirely sure if we want all of these. Happy to change, just wasn't sure what is best


C-Loftus:
#17 I think this addresses this issue. Let me know if you prefer other settings Pokey

pokey (approved):
Ahh much better 😄

PR #22, opened 2024-02-15
 add text to speech functionality

C-Loftus:
One of the biggest issues with asking questions to the model is the fact that it has to copy and paste the response or open up a window to visually show you the response in a buffer somehow. However, if you use text-to-speech, then you don't have to switch any windows around, and this will reduce the amount of clicks that you have to do.

This is dependent upon the text-to-speech functionality in the Sight Free Talon repository. Text-to-speech is deceptively annoying to implement, and as a result, it is better to have it in its separate repository rather than duplicating all the code here.

That repository is generally stable as a dependency. The only thing to note is the fact that, by default, the repository adds echo back for dictation, which users will have to disable through a Talon setting if they don't want it. So, this is not a hard thing to do, but it isn't just a one-click install or something analogous.

I originally had this within the sight free repository, but I think this is something that would benefit from having more users.


pokey (approved):
this looks good. I haven't tested it locally fwiw. I also wonder if we want to just use a setting, so that the user doesn't need separate commands for TTS

C-Loftus:
Ok, will merge. Just for context, I didn't want generalized tts since generalized tts could be all provided by the sight-free repo. This was explicitly for commands that the user wants to do nothing but speak out loud, and not type/paste/function call anything. (i.e. commands that act like an assistant)
To speak the output of `model fix grammar` or something analogous, that could be done with no extra code by just using `sight-free-talon`

Issue #11, opened 2024-01-31
Configure OpenAI endpoint


douglasg14b:
There are many tools/services that provide "OpenAI Compatable" endpoints in a bring-your-own-ai sort of way. Even moreso in corporate environments where we may have GPT-4 completion access that don't share or store corporate data provided on different endpoints (ie. From Azure).

Can the OpenAI endpoint be configurable so it transparently "just works" with any OpenAI-compatable endpoint?

C-Loftus:
Yes that is definitely feasible and probably relatively easy.  if you want to submit a PR, you could just create a setting to change the following from a string to a talon setting https://github.com/C-Loftus/talon-ai-tools/blob/62a1ad117ef5433462eb6dddb59bd6150683688f/GPT/gpt.py#L64  and then the user can specify the Talon setting for the url in a talon file

 only thing I will say is sometimes the default system prompt is a bit different,  for instance for llamafiles I have to specify ( I don't know if this is a similar thing with azure I do not use that at all)

```
'messages': [
                    {
                        "role": "system",
                        "content": "You are an assistant helping an office worker to be more productive."
                    },
                    {
                        'role': 'user', 
                        'content': f"{prompt}:\n{content}"
      }
```

douglasg14b:
Thanks for the reply!!

For Azure at least they provide direct access to OpenAI models for enterprises. So it's the same model.

Is there any weirdness if this points at an endpoint of that uses GPT-4?

C-Loftus:
I don't think so, but I'm honestly not sure since I haven't tested much with GPT-4 in general. (For personal use, a lot of people don't get access to it unless they have paid over a certain threshold, which my usage is not at yet.)

If you create a pull request to support this, I would be happy to merge. That being said, I will probably not implement this on my own since I don't have any endpoints to test with.

This is generally a good idea though and would help reduce some of the duplicate code I have for the llamafile anyways.

C-Loftus:
@douglasg14b  I believe I generalized this to support Azure and any compatible endpoint. Can you check out https://github.com/C-Loftus/talon-ai-tools/pull/24 and see if it works with your azure endpoint?  (you can use the gh cli tool to do this very easy if you aren't familiar)  If you can't test that is okay, since I tested and it works with local llamafiles, so at the very least that pull request is not a regression

C-Loftus:
Closing since I think this is addressed. Will reopen if we get the chance to test and find it doesn't work for some reason

PR #24, opened 2024-02-18
Support other endpoints and generalize settings

C-Loftus:
This pull request makes it so the endpoint and general settings used for the model are easier to customize.  Addresses https://github.com/C-Loftus/talon-ai-tools/issues/11 

 Also made some the documentation in the readme better and more reflective of the current state of the repository. 

 Only a thing I'm not entirely sure on as if we want a different default system prompt. For OpenAI, you do not have to specify the system prompt and then there's some sort of implicit system prompt used in the background.  I don't think there's a way to know what that is.  But for some models like llamafiles,  you need to explicitly state that so i picked a system prompt that generally seems fine.

### DIFF
--- a/GPT/beta-commands/gpt-function-calling.py
+++ b/GPT/beta-commands/gpt-function-calling.py
@@ -5,7 +5,6 @@ from typing import Any, Optional
 import requests
 from talon import Module, actions, settings
 
-from ..gpt import notify
 from ..lib.types import ChatCompletionResponse, InsertOption, Message
 from .gpt_callables import (
     display_response,
@@ -14,6 +13,7 @@ from .gpt_callables import (
     notify_user,
     search_for_command,
 )
+from ..lib.gpt_helpers import notify, generate_payload
 
 mod = Module()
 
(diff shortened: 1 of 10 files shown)


C-Loftus:
Blocked until Douglas has a chance to look at this (or anyone else with an Azure endpoint)

douglasg14b:
Waiting on access to endpoints to test it out!

C-Loftus:
I think I am going to merge this, @jaresty do you have any opinions? It doesn't break anything so it is not a regression. I basically add a new setting for the endpoint

C-Loftus:
https://github.com/C-Loftus/talon-ai-tools/blob/f6833cf0cfefe5d4c533e5d4e4e112d44b8fcfd8/GPT/beta-commands/gpt-function-calling.py#L45
I'm continuing to refactor some things as well to reduce code duplication. @jaresty  is there a reason we have the code language within a system prompt?  I believe that should be passed in within the user prompt,  unless I'm mistaking something with how it wants the language specification be passed in.

jaresty:
I chose to put the language in the system prompt because I considered it to be a kind of metadata about the context in which the command was being called. I think that more information really should be available as part of the system prompt too. Putting everything into the user prompt requires the LLM to parse the metadata-information it needs to respond to the query from the prompt itself. I think using the system prompt to do so makes it easier on the LLM. 

Also, with please the user can specify natural language that results in insert being called, so it really helps to not have to specify the language in your natural language request.

jaresty:
I believe you can have multiple system prompts, fwiw. Any particular reason you wanted to remove the language?

C-Loftus:
I agree that we want to keep this somehow. Unless I'm overlooking something we can't have multiple system prompts https://platform.openai.com/docs/api-reference/chat/create

Essentially my problem was the fact that the language  is often going to be none if we're not in a code editor and  even if we are in the code editor a lot of our commands could feasibly be unrelated to the language. 

That being said you unfortunately Python fn docstrings need to be string literals and can't be interpolated so I think what I will do is just have a ternary operator to add the code language into the system prompt at the end, but only if it is defined. This isn't perfect and we will still be sending the code language a lot of times when it isn't really needed,  but I suppose it's  not likely to be a big deal.

```
>>> talon.actions.code.language()
''
```

C-Loftus:
added a ternary expression for the language context. I think I am happy with that. So we can add the contextual info like the language you want but it's not always going to add it.  in general there's probably a lot of potential to put contextual info within the system prompt but keeping it pretty simple for now with just the language.  let me know if that last commit above looks good

C-Loftus:
@jaresty if this looks good to you, feel free to approve and merge, I have done a fair bit of testing with both the function calling and standard behavior  so there shouldn't be any regressions here

jaresty:
I haven't had too much time to try it out but I'm ok with rolling forward. We can always change it again.

jaresty (approved):

PR #26, opened 2024-03-02
Code Interpreter functionality

C-Loftus:
Would like to have functionality such that the user can automatically run queries against files on their desktop if they desire. 

Currently WIP,  since I don't really understand the API. Have to look into this more. If I run the current code is responds with an ID for an assistant object but the docs don't show how to get the resulting data from it 

- [ ] Figure out why it is returning None and I can't see the resulting file


C-Loftus:
Closing this since I think dealing with file uploads is out of scope right now

PR #36, opened 2024-03-08
Include link to OpenAI signup

brollin:
Just something I think would be handy.


C-Loftus:
Thanks this is great made one super minor edit.

PR #41, opened 2024-03-17
Adds a model format loop command

jaresty:
If you input some desired behavior and then run model for a loop on it will try to generate loop that solves the problem.

### DIFF
--- a/GPT/staticPrompt.talon-list
+++ b/GPT/staticPrompt.talon-list
@@ -14,7 +14,9 @@ format table: The following markdown text is raw data. There is no index. Return
 format bullets: Convert the every paragraph into a one heading with series of bullet points underneath it. Each paragraph is separated by a new line. Separate paragraphs should not have combined bullet points. This should all be done in markdown syntax. If it is a small paragraph than you can just leave it as a heading and not add bullet points. Do not reduce content, only reduce things that would be redundant. These bullet points should be in a useful format for notes for those who want to quickly look at it. If there is a citation in the markdown original then keep the citation just at the top and not within every individual bullet point.
 format mermaid: convert the following plain text into the text syntax for a mermaid diagram.
 format gant: convert the following plain text into the text syntax for a gant chart within mermaid.
-format loop: the following plaintext describes an iterative process in code in the language that is specified by the system prompt. Please output the code necessary to do this transformation.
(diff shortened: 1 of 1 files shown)


C-Loftus:
I think we should probably change this to `format code` or something.  Since changing selected text into code is not specific to iteration or looping.

jaresty:
I do think it would be good to provide some additional programming constructs; I think providing the context that the expected output is a loop helps the LLM to give a more targeted response. I'm not sure if we need to do all constructs, but conditional was one I was thinking would be good (switch is one kind of conditional too)

jaresty:
A general purpose code one could be good too, though I'm not sure it will do so well at that beyond basic syntax.

jaresty:
Here's a thought-what if we add both and try them out for a while, and if we discover that format loop and format code are roughly equivalent we could deprecate format loop? That way we can test them out before making a final decision there.

jaresty:
That said, I do think there's also value in providing some mechanism for users to customize their own prompts too, which I think is what you were hinting at. I don't think the two things are tied together though.

C-Loftus:
I would prefer to first merge a `model convert code` since I think this is relatively niche. GitHub copilot is really much better at this given the fact it can insert into the proper destinations and gain the context of your codebase, and you don't need to select text beforehand.  I think `format code` is ok since it is not too specific and we already have `shell`. 

It's totally fine if you find `format loop` to be useful. I just think it is too specific to your setup to merge, since I don't think many people write the instructions down before they generate the code. I would prefer to encourage people to use the copilot integration. I know you said you don't use copilot, which is totally fair, but I think there are some open source alternatives that can be used instead. 

It's possible that what we want to do is to change the static list into a talon capture, such that the capture is a logical OR of the prompts included within the static list or the user's special custom prompt list.

jaresty:
We discussed on the phone and I think we arrived at a good compromise:

- move these prompts to a new section called "code generation"
- put that towards the bottom of the file
- use a different namespace "generate" to prefix the commands

This way it doesn't collide with the model format commands and makes IT fairly simple to see and experiment with the code generation ones.

C-Loftus:
Looks good thanks. If the underlying model changes affect this at all going forward let me know and we can discuss going forward in more depth about pinning  model versions

PR #52, opened 2024-04-14
Support for Different Shells

C-Loftus:
Supports a setting to specify the shell name. 

@jaresty let me know if this is what you want. Nushell probably will not work particularly well since the model has less context for it. But GPT 4 might work better.

Either way this is a useful setting since I sometimes I need powershell if my terminal is windows terminal instead of my usual Linux setup


jaresty:
This looks great!

PR #53, opened 2024-04-14
Custom prompts and easier customization

C-Loftus:
Instead of using a raw talon list, we now use a capture `user.modelPrompt` for all the model commands. The capture is formed via the set union of static prompts and custom prompts. Custom prompts are empty by default. There is an example that can be copied into the user directory from the GPT folder.

The user can create a custom prompt talon list file outside this repo in their user directory and it will automatically be the most specific  (i.e. you don't need to do weird context matching on the hostname or stuff that is semantically meaningless)


C-Loftus:
@jaresty when you get a chance could you try this out locally and see if you like the UX? Also curious on the naming scheme and if you think customPrompt and modelPrompt both make sense. (as well as static prompt still). 

`modelPrompt = staticPrompt U customPrompt` <- does that seem fair if you had no prior background with this repo? 

If it seems reasonable I will document this in the readme for `GPT/` and then merge

jaresty:
This is great! 😊 I really like the new interface and I think I can use it easily to experiment with new prompts. Thank you for putting this together! 🌟

jaresty (approved):

C-Loftus:
Sweet will doc this then merge. Just to confirm, did you have a chance to test locally? If not it is ok, just want to know for context

jaresty:
I did give this a try locally and it worked for me.

PR #55, opened 2024-04-16
Sql support

C-Loftus:
Essentially model shell but for sql


C-Loftus:
@jaresty opinions on this? I don't write that much sql so wasn't sure

jaresty:
I think this could be pretty useful, but I'm pretty sure you'd need to pass in at least a database schema for it to produce valuable outputs. Have you tried it?

jaresty:
Could you share some example prompts and responses? I think it would be helpful to get a sense of how to use it. I'm curious enough to give it a try.

C-Loftus:
Sure and no pressure obviously, just was curious since I value your perspective.

I think as well it would be relatively straightforward to get the schema and pass it into the prompt to make sure the syntax is valid, but that's once again probably out of scope for the PR

> model sequel what is my database version

> model sequel get the counts of all names in the user table

> model sequel generate a schema for table that represents a city in a geospatial application

Once again this is stuff that is moreso used for educating/fixing syntax/pair programming and since it uses the confirmation GUI doesn't run anything automatically

jaresty:
Those seem like practical use cases. I'm wondering if there's a concept missing from the gpt usage: a way to pin some context to be sent with a query. 

I could imagine a usage similar to how rango marks work. You could name a mark, add to it, and then send it to ChatGPT by name. Something like this:
- model add block X to mark brainstorm 
- model add file X to mark brainstorm
- model process mark brainstorm with <prompt>

Or something like that that allows you to add the schema to the context before sending it.

PR #56, opened 2024-04-19
Add helper to run a query and select the response

jaresty:


GPT/gpt.py
```diff
@@ -110,6 +110,14 @@ def paste_model_confirmation_gui():
         GuiState.text_to_confirm = ""
         confirmation_gui.hide()
 
+    def paste_selected(result: str):
```
C-Loftus, line 113: Isn't `paste_and_select` perhaps more clear?

C-Loftus:
This looks like a good idea. I like how it mimics statefulness without needing to explicitly add state internally. I made a few comments then I think we can merge?

GPT/gpt.py
jaresty: Works for me 👍

GPT/gpt.talon
```diff
@@ -15,6 +15,12 @@ model clip <user.modelPrompt> [this]$:
     result = user.gpt_apply_prompt(modelPrompt, text)
     clip.set_text(result)
 
+# Runs a model prompt on the selected text and sets the selection to the pasted response
+model selected <user.modelPrompt> [this]$:
```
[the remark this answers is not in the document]
jaresty: Sounds good. Changed ✅

C-Loftus:
Thanks for making edits. I will test this out locally later/tomorrow, perhaps add some docs, change the old clip command, then merge

C-Loftus:
Tested. Merging this! Thanks for this idea, I think it will be useful! I am curious to continue thinking about other results / pipelines for model outputs (i.e. we have select, clip, and I wonder if there are other ways to process that would be useful. Will continue to brainstorm for future PRs)

PR #57, opened 2024-04-20
Add a helper for when you want to append the response

jaresty:


GPT/gpt.py
```diff
@@ -118,6 +118,12 @@ def paste_and_select(result: str):
         for _ in result:
             actions.edit.extend_left()
 
+    def paste_and_append(result: str):
```
C-Loftus, line 121: Honestly not sure if we need this function. I think it makes sense to keep stuff in the `.talon` file if it is just talon actions without any Python code. I like to limit stuff in the global namespace if possible. Is that reasonable?

jaresty: That makes sense to me! ✅

GPT/gpt.talon
```diff
@@ -21,6 +21,13 @@ model <user.modelPrompt> [this] (then | and) select [it]$:
     result = user.gpt_apply_prompt(modelPrompt, text)
     user.paste_and_select(result)
 
+# Runs a model prompt on the selected text and appends the result to the next line
+model <user.modelPrompt> [this] (then | and) append [it]$:
```
pokey, line 25: I'm not sure about the spoken form here. For one I am skeptical of all the optional words. We originally went this route in cursorless and then shortened things dramatically to one consistent grammar. I am of the mind that adding support for lots of optional words is pretending we understand English, which will lead to problems long term, given that we don't

But even accepting the optional words (given that is somewhat consistent with other commands in this file), I don't think the `then | and` makes sense here, because to me it implies the selected text will get updated and then there will be an append, similar to how the command above modifies selected text and then reselects it

Maybe something like
 
```suggestion
model <user.modelPrompt> after$:
```

alternatives to `after`:

- `down`
- `append`
- `clone`

we could consider `model append <prompt>` as well

pokey (approved):

GPT/gpt.talon
pokey: Would be helpful to see a couple example prompts where that is awkward

GPT/gpt.talon
```diff
@@ -1,32 +1,20 @@
-# Ask a question in the voice command and the AI will answer it.
+# Ask a question in the voice command and the model will answer it.
 model ask <user.text>$:
     result = user.gpt_answer_question(text)
     user.paste(result)
 
-# Runs a model prompt on the selected text and pastes the result.
-model <user.modelPrompt> [this]$:
+# Runs a model prompt on the selected text; pastes or inserts via the specified method
+model <user.modelPrompt> [this] [{user.modelInsertionMethod}]$:
```
C-Loftus, line 7: Changed the talon syntax to always support an optional insertion method. How does this look?

GPT/modelInsertionMethod.talon-list
```diff
@@ -0,0 +1,10 @@
+list:  user.modelInsertionMethod
+
+-
+
+above
+below
+clipped
+selected
```
C-Loftus, line 8: Let me know if these seem reasonable. There is always a tradeoff in wording since some of these are locations and some of them are moreso actions (i.e. selecting), but I was hoping to have them all fit syntactically in the same commandgrammar. 

I am using 'above' and 'below' since 'after' in my opinion isn't as clear since it could imply something temporal.

jaresty:
Looks good to me!

pokey:
Looks like talonfmt doesn't support python 3.12. Might have a look at that repo to see why that's pinned

C-Loftus:
Yup I did see that. But I didn't change Python versions or update talonfmt so I am confused why this is giving an error now. This is causing PRs in community to fail now as well. Don't see any upstream reason why. I think black got bumped in the auto formatter to use 3.12 and I guess now that makes the precommit use python 3.12 so it fails or something?

Just going to wait on this since it isn't super clear to me what is going on

PR #59, opened 2024-04-29
don't try to use ~/Downloads if it doesn't exist

fidgetingbits:
ATM things like `model help` fail if `~/Downloads` doesn't exist, so this fixes things to just use `/tmp` if `~/Downloads` doesn't exist. For snaps, which I don't use, I'm not sure how they deal with that folder not existing. Perhaps they use `/etc/xdg/user-dirs.defaults`, but not worth getting too fancy until it's needed.


C-Loftus:
Ok this looks good thanks! :+1: Yeah unfortunately not really sure the best way to deal with the snap issue without doing some subprocess hack.

PR #68, opened 2024-06-16
use cursorless destinations instead of pasting

jaresty:


C-Loftus:
Oh awesome, is this part of the public API now? iirc this wasn't possible previously

jaresty:
Yes! It is now 😀

PR #69, opened 2024-06-16
add a smart clipboard command

jaresty:


pokey:
Can you give an example of how you'd use this?

jaresty:
Here's a concrete example:

Write a comment describing what you want to change, or pseudo code a type definition (etc), then "model bring" some similar code to that spot and the LLM should make the changes you need.

C-Loftus:
This is cool and thanks for submitting. I just feel that in the context of that original article, they had more of a need given the fact that they are not using cursorless, and things had a bit less friction since they aren't using voice commands. When both of those come into play, I wasn't sure if that would decrease utility. I am skeptical if the LLM has enough context to intelligently paste.

>  Write a comment describing what you want to change, or pseudo code a type definition (etc), then "model bring" some similar code to that spot and the LLM should make the changes you need.

Is it possible already do similar things using `model please` and/or `model generate code` on a cursorless target?

jaresty:
You can do the same but it is more verbose. This is very succinct. You can, for example, just write a single word and then use bring or paste to get the LLM to infer the transformation you intend. No worries if you don't want to accept the PR-I'm happy to just keep this for my own use instead, but I do think it has utility and I'm interested to experiment and see how it is used.

jaresty:
Fwiw - I like to lean on LLMs to give feedback as well. When you aren't exactly sure how you want to change a thing, using "models please" would be harder than "model paste," imo. You can model paste to a block and it will try to figure out the best place to insert what you're trying to do.

jaresty:
With respect to the amount of context-it doesn't need too much from the code base to be useful. "Model please" (and other commands) operate on limited context mostly drawing from the general corpus available to the LLM. This is basically doing the same but with the minor difference that it is intended to combine two chunks of text instead of modifying just one. (There is also a more general version of this where you can specify how you want them to be combined, but I didn't go that far with this)

GPT/gpt.py
```diff
@@ -61,6 +61,20 @@ def gpt_answer_question(text_to_process: str) -> str:
         """
         return gpt_query(prompt, text_to_process)
 
+    def gpt_smart_clipboard(clipboard_text: str, destination_text: str):
```
C-Loftus, line 64: I think this could be renamed `gpt_smart_combine` or something analogous just to be super clear since it isn't really using the clipboard here, even if the prompt refers to it.

C-Loftus:
That's all fair. I am fine with the idea. I think it makes sense and I agree that `model please` or `model generate code` can sometimes add extra friction when chaining them together.

I do think that we might need to think a bit about the API. `model paste` is not super intuitive and the python function is acting like a clipboard, but isn't really using it. 

I think this is  worth merging but I'd just like to keep this in the gpt-beta folder and change a few things first to make it super clear if that is fair with you.

jaresty:
I'm definitely open to suggestions on changing it, and also fine with putting it in beta for now 😀

pokey:
Oh I get it now. Interesting idea. Is it working well in practice?

jaresty:
I'd say it's still experimental. I think it makes sense to put it into the beta flag-it works pretty well but I think I might want to tweak the prompt a bit and also I'm still figuring out how to use it. It feels pretty fun to use (experimental/kind of like play). I do think it's worth trying out.

GPT/gpt.talon
```diff
@@ -11,6 +11,12 @@ model <user.modelPrompt> [this] [{user.modelInsertionMethod}]$:
     result = user.gpt_apply_prompt(modelPrompt, text)
     user.gpt_insert_response(result, modelInsertionMethod or "")
 
+model paste:
```
C-Loftus, line 14: Perhaps we can think if there is a different way to word this. Since `model paste` might be confusing and is a bit vague. I am not really sure a better alternative though. 

`model combine` is a bit more semantic I feel. But not super happy with that either

C-Loftus: `model combination paste` is long but is very clear. I would be fine with that honestly.

GPT/gpt-cursorless.talon
```diff
@@ -9,3 +9,10 @@ model <user.modelPrompt> <user.cursorless_target> [<user.cursorless_destination>
     result = user.gpt_apply_prompt(user.modelPrompt, text)
     default_destination = user.cursorless_create_destination(cursorless_target)
     user.cursorless_insert(cursorless_destination or default_destination, result)
+
+model bring <user.cursorless_target> to <user.cursorless_target>$:
```
C-Loftus, line 13: Almost wonder if this should be named `model merge`. But I understand wanting to keep cursorless syntax too.

C-Loftus (commented):
Generally looks fine and I think its an interesting long term idea we might as well have in the repo even as a stretch goal. 

Perhaps you can think on this a little bit, then make any changes you see fit (or say disagree and not change anything you don't want, I just think the API is just a bit confusing for new users), move it into GPT-beta, and then we can merge.

GPT/gpt.talon
jaresty: What about "model blend paste"? For the clipboard version and "model blend <cursorless destination >" for the cursorless flavor?

jaresty: Or maybe even just "model blend [target] [destination] where:

1. If no target or destination it blends with clipboard
2. If target and no destination, it blends every target and pastes the result
3. If target and destination then it blends every target and the destination and inserts to the destination

C-Loftus: I like the latter with `model blend [target] [destination]`. Sounds good to me and seems pretty extensible! 

with that in mind, I think we should rename the `gpt_smart_clipboard` function. Since we aren't really using the clipboard or pasting.

jaresty: Yep, makes sense. I'll change that up shortly.

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -14,3 +14,8 @@ app: vscode
     text_list = user.cursorless_get_text_list(cursorless_target)
     default_destination = user.cursorless_create_destination(cursorless_target)
     user.gpt_dynamic_request_cursorless(user.text, text_list, cursorless_destination or default_destination)
+
+model blend <user.cursorless_target> [<user.cursorless_destination>]$:
+    target_text = user.cursorless_get_text_list(cursorless_target)
+    result = user.model_blend(target_text, edit.selected_text())
+    user.cursorless_or_paste_helper(cursorless_destination or 0, result)
```
pokey, line 21: This doesn't look right to me. If you supply a destination, it uses selected text in prompt context, but then places the output at destination. I had thought the idea was for destination to be used for both prompt context and output if destination is supplied

jaresty: Yeah, you're right. I was having terrible getting the text from the cursorless destination. Is there an api for that? The alternative could be to use a cursorless target for the destination and use the api to make that into a destination, which should work. What do you think?

pokey: Yeah I think I'd do that for now and only support "to"

jaresty: Sounds good. I'll plan to do that next chance I get.

jaresty:
I think this is ready to go now.

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -14,3 +14,22 @@ app: vscode
     text_list = user.cursorless_get_text_list(cursorless_target)
     default_destination = user.cursorless_create_destination(cursorless_target)
     user.gpt_dynamic_request_cursorless(user.text, text_list, cursorless_destination or default_destination)
+
+model blend <user.cursorless_target> to <user.cursorless_target>$:
+    target_text = user.cursorless_get_text_list(cursorless_target_1)
```
pokey, line 19:
```suggestion
    target_text = user.cursorless_get_text(cursorless_target_1)
```

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -14,3 +14,22 @@ app: vscode
...
+
+model blend to <user.cursorless_target>$:
+    target_text = edit.selected_text()
+    destination_text = user.cursorless_get_text(cursorless_target)
+    default_destination = user.cursorless_create_destination(cursorless_target)
+    result = user.model_blend(target_text, destination_text)
+    user.cursorless_insert(default_destination, result)
+
+model blend <user.cursorless_target>$:
+    target_text = user.cursorless_get_text_list(cursorless_target)
```
pokey, line 33:
```suggestion
    target_text = user.cursorless_get_text(cursorless_target)
```

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -14,3 +14,22 @@ app: vscode
...
+    result = user.model_blend(target_text, edit.selected_text())
+    user.cursorless_or_paste_helper(cursorless_destination or 0, result)
```
pokey, line 35: Destination will never be defined here

GPT/beta-commands/beta-cursorless.talon
jaresty: I was on the fence about using the singular version because I've found that the list call seems to work just fine here but also allows you to specify multiple targets (where the singular api call throws an error). Is this a better choice?

jaresty:
@pokey i left an open question on the PR. I think we're just waiting on your feedback now.

GPT/beta-commands/beta-cursorless.talon
pokey: That's fine you should just update the type hint for the first arg on `model_blend` to indicate it's a list of strings, not a string. Tbh I'm surprised talon doesn't complain about the incorrect runtime type there

jaresty: Is there a way to create a list of string in talon script? For the clipboard usage I was struggling with how to do that.

jaresty: I think I'll make another function for the list use case. I can provide more context to the LLM that way anyhow.

jaresty:
I extracted a new helper for dealing with lists.

GPT/gpt.py
```diff
@@ -65,6 +78,16 @@ def gpt_answer_question(text_to_process: str) -> str:
...
 
+    def model_blend(source_text: str, destination_text: str):
+        """Blend all the source text and send it to the destination"""
+
+        return model_blend(source_text, destination_text)
+
+    def model_blend_list(source_text: list[str], destination_text: str):
+        """Blend all the source text as a list and send it to the destination"""
+
+        return model_blend("\n---\n".join(source_text), destination_text)
```
pokey, line 89: Is this working better than the original version which just used Python's built in list to string conversion?

GPT/gpt.talon
```diff
@@ -11,6 +11,12 @@ model <user.modelPrompt> [this] [{user.modelInsertionMethod}]$:
     result = user.gpt_apply_prompt(modelPrompt, text)
     user.gpt_insert_response(result, modelInsertionMethod or "")
 
+model blend:
```
pokey, line 14: Not sure bout this spoken form. Would be tempted to use something like

- model blend clip
- model paste
- model paste blend

I don't feel strongly tho so if you've already aligned on this one then feel free to ship this spoken form

GPT/gpt.py
jaresty: It's hard to say for certain, but I think so. It does give a little bit more context to the LLM to aid with how to handle multiple sources.

GPT/gpt.talon
jaresty: I'm ok with `model blend clip`. I was starting to get used to the `model blend` so I might have a little bit of interference but I think that's fine.

jaresty: Updated it

pokey:
Ok looking good. My last question is why everything is anchored (the final `$` on each command). I generally only anchor things that end with a prose capture. Any reason we shouldn't be able to chain these with other commands?

jaresty:
No reason I can think of. I think I just copied that from somewhere-will fix.

GPT/beta-commands/beta-cursorless.talon
```diff
@@ -14,3 +14,21 @@ app: vscode
...
+    destination_text = user.cursorless_get_text(cursorless_target_2)
+    default_destination = user.cursorless_create_destination(cursorless_target_2)
+    result = user.model_blend_list(target_text, destination_text)
+    user.cursorless_insert(default_destination, result)
+
+model blend to <user.cursorless_target>:
+    target_text = edit.selected_text()
+    destination_text = user.cursorless_get_text(cursorless_target)
+    default_destination = user.cursorless_create_destination(cursorless_target)
+    result = user.model_blend(target_text, destination_text) * user.cursorless_insert(default_destination, result)
```
pokey, line 29: What's going on with this line?

pokey (commented):
Found one more issue

I do think these could all be unified into one command, but let's leave that for a follow-up. This PR is already getting long in the tooth

GPT/beta-commands/beta-cursorless.talon
jaresty: Looks like a typo that got reformatted automatically. Just fixed it-thanks!

pokey (approved):
Ok I'm happy. In the future we can refactor to merge these implementations

C-Loftus:
@jaresty I did a final refactor just to clean up a few things.  You can merge once you are satisfied. (Perhaps just give this a test run for a day or so to make sure I didn't regress anything)

Just as a note, I put any `model blend` commands into the beta folder and added a few comments as needed. I still think the API is still a bit tricky to intuit but we can iterate on it as we use it more.

jaresty:
Fixed a small bug but now I'm going to merge it! thanks all!!!

PR #71, opened 2024-06-25
 Strip markdown that is output despite our prompting 

C-Loftus:
Sometimes with the model responses, even though we explicitly tell it not to output markdown syntax, it still does it. This PR uses a regex to remove the wrapper.

It also implements pytest since this is just a pure function and thus can be tested more easily.


C-Loftus:
@jaresty I think this should implement the markdown strip we were looking for.  Would you mind testing this out a bit and see if you run into the markdown issue? I am no longer getting it but wasn't sure your context or if specific commands were reliably causing markdown syntax and overriding the prompt.

Ignore the branch name, there was a rebase issue but it is fine now and won't be an issue when this is merged.

jaresty:
I'm happy to work off this one-do you think we could integrate the blend changes here? I'm using that one a bit so that's holding me back from switching to this branch.

C-Loftus:
> I'm happy to work off this one-do you think we could integrate the blend changes here? I'm using that one a bit so that's holding me back from switching to this branch.

Hmmm we can although it might just be easier to try doing `gh pr checkout 71` and then testing a few things quickly. I think the other PR is blocked a bit and has to be tidied up a bit. However, if blend was causing the markdown output, then yeah perhaps this is blocked on the blend PR.

jaresty:
I gave it a shot on a few of my prompts and I did not see the markdown code blocks so it seems to be working based on my cursory checking right now.

C-Loftus:
Ok, will merge this and then we can get the other pr branch up to date. Shouldn't be any regressions but if so I can just revert.

PR #79, opened 2024-06-30
Paste and select can be much faster

jaresty:
- First iterate through all new lines and extend upward
- Then select just the first line


jaresty:
Speeding this up makes it a lot more usable.

C-Loftus:
This is great, thanks!

PR #81, opened 2024-06-30
Cleanup model blend cursorless grammar

jaresty:
- Model blend to <target> didn't specify a source; now it specifies the source as clipboard
- Model blend <target> was only implicitly dropping the 'this'

### DIFF
--- a/GPT/beta-commands/beta-cursorless.talon
+++ b/GPT/beta-commands/beta-cursorless.talon
@@ -20,7 +20,7 @@ model blend <user.cursorless_target> to <user.cursorless_target>:
     result = user.gpt_blend_list(target_text, destination_text)
     user.cursorless_insert(default_destination, result)
 
-model blend clip to <user.cursorless_target>:
+model (blend clip | paste) to <user.cursorless_target>:
     clipboard_text = clip.text()
     destination_text = user.cursorless_get_text(cursorless_target)
     default_destination = user.cursorless_create_destination(cursorless_target)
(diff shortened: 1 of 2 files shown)


pokey:
This seems consistent. I think for convenience it would also be good to have:

- "model paste" or "model paste that"
- "model paste to <target>"

Maybe we already have these. But they would be pretty consistent with "paste that" and "paste to <target>"

jaresty:
I like those. Updated.

C-Loftus (changes requested):
I think we should either use `clip` or `paste` and not have both in the capture. It makes it a bit cleaner to talk about commands when there is one canonical representation for any keywords inside of it.

Let's try to keep non-pipelined commands in a form somewhat similar to 
`model $prompt $textSource $responseMethod` since it makes the mental model easier for the user when everything follows a similar grammar.

So `model blend clip to $cursorless_target` in my opinion is more consistent with this grammar if you are saying it is blending the clipboard as the text source with the cursorless target as the insertion method.

jaresty:
I'm ok with either of these. Can you discuss at the meetup and make a call which way to go? I'll be afk today but I'm ok if you want to make whatever change you agree to and merge it.

jaresty:
If you word prefer I change it, just let me know what to do and I'll put it together tonight if I can find some time. Please feel free to edit it though. The current cursorless blend command is missing a clipboard option so wanted to fix that up.

jaresty:
I think this aligns with the current grammar. I wasn't sure if I should be adding something for 'paste'.

C-Loftus (approved):

C-Loftus:
Looks good thanks!

PR #84, opened 2024-07-08
Adds a snippet insertion destination

jaresty:
- This is especially useful when you are writing code


pokey:
Very cool idea. How well does it work in practice?

I'm not sure the fact of it being a snippet is a destination; seems somewhat orthogonal but I'm not sure I have a much better idea and this stuff is still all quite experimental so probably ok

jaresty:
It seemed to work pretty well when I was testing last night. I'm going to keep trying it out and see how it goes. I'm excited at the possibility it provides for the LLM to clearly point out the parts of the response that should be adjusted, which is a pretty useful way to communicate.

jaresty:
I agree about it being somewhat orthogonal to insertion method-it's more of an adverb that could be applied to above/below/replace. We don't have that in the grammar yet though.

C-Loftus:
Interesting idea. I feel like this could significantly benefit from multishot prompting. Since I feel like the model could interpret the prompt as it is now in many different ways. But multishot prompting is probably a story for another PR.

If you are finding this to be practical, I think we should create a new talon list called `modelInsertionModifier` or something similar. We can then also include `selected` in that list if you want to always paste with text selected. I would prefer to pass in `modelInsertionModifier` instead of `modelDestination` just to make it very clear that `snippet` is not a destination, but a way by which the text is inserted differently.

But I'd experiment with things a bit before making too many changes.

C-Loftus:
And once again my only hesitancy with this is that is does seem very experimental and I am not really sure if the added complexity to the grammar is worth it. I think most people would want to just say the snippet in community they want. Arbitrary text instructions usually doesn't conform to snippets as well compared to well defined blocks (classes, if statements, specific boilerplate etc)
As with everything this does have a use, I suppose we just have to see how useful since I haven't played around with this myself.

jaresty:
`insertionModifier` sounds good to me

GPT/gpt.py
```diff
@@ -192,20 +195,32 @@ def gpt_reformat_last(how_to_reformat: str):
             notify("No text to reformat")
             raise Exception("No text to reformat")
 
+    def gpt_paste_with_modifier(text: str, modifier: str = ""):
```
pokey, line 198: I think I would just call this `gpt_insert`. And I would be tempted to just have `modifier` be a boolean `is_snippet` for now as it's unclear we'll actually need more modes, and the name `modifier` is quite vague

GPT/gpt.py
```diff
@@ -192,20 +195,32 @@ def gpt_reformat_last(how_to_reformat: str):
...
-        result: str, method: str = "", cursorless_destination: Any = None
+        result: str,
+        modifier: str = "",
+        pasteDestination: str = "",
+        method: str = "",
+        cursorless_destination: Any = None,
     ):
         """Insert a GPT result in a specified way"""
-        match method:
+        match method or pasteDestination:
```
pokey, line 214: this confuses me. Why are we switching on both these variables?

GPT/gpt.talon
```diff
@@ -5,10 +5,10 @@ model help$: user.gpt_help()
 #   Example: `model fix grammar below` -> Fixes the grammar of the selected text and pastes below
 #   Example: `model explain this` -> Explains the selected text and pastes in place
 #   Example: `model fix grammar clip to browser` -> Fixes the grammar of the text on the clipboard and opens in browser`
-model <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
+model <user.modelPrompt> [{user.modelSource}] ([{user.modelPasteDestination}] [{user.modelInsertionModifier}] | [{user.modelDestination}]):
```
pokey, line 8: what is the difference between a paste destination, an insertion modifier, and a model destination?

pokey (commented):
I think this is def worth having at least in experimental form, as it's a really cool idea and I could see the model generating some interesting snippets on the fly

I find the code / terminology a bit confusing tho; see comments inline

Also, total nitpick, but there is lots of camelCase in this repo, whereas it is conventional to use snake_case in talon / python

GPT/gpt.talon
C-Loftus: Yeah I don't particularly want to have an additional list for paste destination in addition to the model destination. The issue is that whenever we split things into separate lists we have to check more permutations to make sure they are valid (ie we can't paste above in a window or with verbal output) and paste destination doesn't really strike me as semantically different enough to warrant its own list. Unless I am missing something

GPT/gpt.py
jaresty: I split the original talon list into two but didn't split the match statement. The match statement could be split too. Only one of these will be set though.

GPT/gpt.talon
jaresty: I could also see an argument that "below" and "above" are insertion modifiers as well-what if we were to put those in the same talon list and allow multiple to be applied?

GPT/gpt.py
```diff
@@ -192,21 +195,28 @@ def gpt_reformat_last(how_to_reformat: str):
...
-            case "above":
-                actions.key("left")
-                actions.edit.line_insert_up()
-                actions.user.paste(result)
-                GPTState.last_was_pasted = True
-            case "below":
-                actions.key("right")
-                actions.edit.line_insert_down()
-                actions.user.paste(result)
-                GPTState.last_was_pasted = True
```
C-Loftus, line 209: We should keep GPTState info. We still want to keep track of things in case they want to select it after the fact. Most users will select after the fact, not in the request. If we get rid of this variable we don't know if the last response can be pasted.

C-Loftus:
To make things consistent with the other code in this repo, I made a couple quick edits and opened a new PR #85 , let's discuss further there. I actually think that snip should be a prompt with arbitrary text after it like `ask` so I made some refactors.

GPT/lists/modelInsertionModifier.talon-list
```diff
@@ -0,0 +1,17 @@
+list:  user.modelInsertionModifier
+
+-
+
+# A list of adjectives/adverbs describing how the result of the GPT query should be returned
+
+# Inserted as a snippet
+snip: snip
+
+# paste in place
```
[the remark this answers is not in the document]
jaresty: I was thinking of it as kind of like the null modifier-as in: insert this where it is without changing the cursor at all. As this is currently written you cannot combine destinations with modifiers. All the destinations are targets outside of the editor. I don't think that it makes sense to paste a snippet to browser or clipboard for example.

GPT/gpt.py
jaresty: I'm still tracking it below in the default case. It will be set for above and below as well as for snippets. Is that still a problem?

GPT/lists/modelInsertionModifier.talon-list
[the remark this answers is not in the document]
C-Loftus: `to this` follows the pattern of the cursorless grammar. I would prefer to follow that convention. 

It may not feel like a lot of complexity but I am really trying to focus on getting users using anything beyond just the default destination/source. It makes the documentation site a bit nicer if we keep the `source` `destination` pattern.

I agree that it doesn't make much sense to apply `snip` to the clipboard, but it follows the grammar and doesn't cause anything to break. Once we starting allowing multiple modifiers which are mutually exclusive with a destination, it starts to get messy.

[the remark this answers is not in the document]
jaresty: If you'd prefer I structures so that you can apply insertion modifiers to every destination then I can readjust. That is preferable to me. 👍

jaresty:
I made another alternative inspired by #85 : https://github.com/C-Loftus/talon-ai-tools/pull/86

C-Loftus:
Closing in favor of #89

PR #85, opened 2024-07-09
snippet refactor

C-Loftus:
@jaresty made some changes to support snippets in a new PR

-  I think `snip` should be a prompt, not an insertion modifier. 
`model snip generate a pandas dataclass` or `model snip generate a class`

- I think we shouldn't really have an insertion modifier list since it adds more complexity
   -  My number one goal is to improve discoverability right now. I think this hurts that. 
- Kept some things that were removed in the other PR that I mentioned in my last comment.

In general I also am not clear how to use this. How do you have it insert a snippet like talon inserts snippets where you can jump between sections? I don't find the prompt performs that well but maybe I am doing something wrong

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -153,11 +153,10 @@ class UserActions:
             else text_to_process
         )
 
-
         """
         Handle special prompts that require arbitrary text.
 
-        These prompts treat <user.text> as the source by setting 
+        These prompts treat <user.text> as the source by setting
         text_to_process to the natural language request.
         """
 
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -153,11 +153,21 @@ class UserActions:
             else text_to_process
         )
 
-        # Ask is a special case, where the text to process is the prompted question, not the selected text
+
+        """
+        Handle special prompts that require arbitrary text.
+
+        These prompts treat <user.text> as the source by setting 
+        text_to_process to the natural language request.
+        """
+
         if prompt.startswith("ask"):
             text_to_process = prompt.removeprefix("ask")
             prompt = """Generate text that satisfies the question or request given in the input."""
         # If the user is just moving the source to the destination, we don't need to apply a query
+        elif prompt.startswith("snip"):
+            text_to_process = prompt.removeprefix("snip")
+            prompt = "Please return the response as a textmate snippet for insertion into an editor with placeholders that the user should edit. Return just the snippet content - no XML and no heading."
         elif prompt == "pass":
             return text_to_process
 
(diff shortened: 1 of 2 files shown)


jaresty:
I prefer this to be an insertion modifier because it can be used with various kinds of prompts. Having it be its own prompt makes it really limited, since you aren't able to use it with all of the prompts that you already have. The way I was using it seemed to be working really well for me-did you try that PR?

jaresty:
For what its worth that was using it with model please

C-Loftus:
`model snip <user.text>`  in this PR is equivalent to `model please <user.text> snip` . It is the same thing just less verbose.

Regardless, I tried it with the other PR. In both I am not finding it to work well. Does it insert with blanks like  Talon snippet for you? I am finding that it just generates code like it would for a normal model please request. The advantage of a talon snippet in vscode would be the ability to jump between sections but I am not finding this does that.

Both images are results of the other PR. Same sort of results with this PR

`model please generate a dataclass snip`
image

`model please generate a while loop snip`
image

jaresty:
It is generating with placeholders for me: 
image

jaresty:
Did you try using snip next?

jaresty:
I want to use it with model answer, which this grammar would not allow for.

C-Loftus:
For me I now find it generates snippet placeholders around 25% of the time for simple syntax-based requests like generating a for loop. But anything above more complicated doesn't seem to work.

jaresty:
I was getting much higher success rate than that 🤔 and for more complex results as well.

C-Loftus:
Could you give examples?

jaresty:
"model generate a data class snip"
image

jaresty:
Maybe this is because I'm using gpt-4o

jaresty:
`model answer snip` with a code block alone turns it into a snippet that allows me to edit the values that I might want to just using `snip next`

image

jaresty:
I could definitely see adding a `model paste snip` that does the same for you

C-Loftus:
Ok seems to be a gpt 4o difference. Closing this

jaresty:
I could get behind this being a custom prompt like this if it also considered the current selection (like model answer). I think there are two usages:

1. With a text prompt, spoken
2. Operating on the selection or the clipboard

jaresty:
I'm warming up to this direction-can we reopen it?

jaresty:
I think I'd prefer it behave more like a regular prompt though-maybe we can defer solving adding additional context until we do it for other prompts as well.

C-Loftus:
Reopened. I do this generally fits my goals with the format of things better. (i.e. prompts that are easy, follow a grammar that is somewhat similar to cursorless, and have limited amounts of helper fns/special cases, error handlings/extra talon lists)

However,  I am not sure how to make it so you can also use model answer with it. Don't have a lot of bandwidth at the moment to play around with things unfortunately.

jaresty:
Here's another flavor of this that optimizes for the model answer use case and makes this operate more like other prompts: https://github.com/C-Loftus/talon-ai-tools/pull/86

jaresty:
I think what's actually missing here may be that you aren't using the cursorless insertion method to paste unless I missed it. `actions.user.insert_snippet(text)`

PR #86, opened 2024-07-10
Alternative snippet refactor

jaresty:
- This make snippet just like any other prompt
- you can run it on the clipboard or on selected text
- If we someday get around to adding the ability to pass additional context to every prompt, snippets would get it as well

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -199,12 +199,12 @@ class UserActions:
             case "above":
                 actions.key("left")
                 actions.edit.line_insert_up()
-                actions.user.paste(result)
+                actions.user.insert_snippet(result)
                 GPTState.last_was_pasted = True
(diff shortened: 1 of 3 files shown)


GPT/gpt.py
```diff
@@ -200,12 +199,12 @@ def gpt_insert_response(
             case "above":
                 actions.key("left")
                 actions.edit.line_insert_up()
-                actions.user.paste(result)
+                actions.user.insert_snippet(result)
```
jaresty, line 202: insert_snippet falls back to paste, so it should be equivalent for most use cases

jaresty: never mind-it turns out that it's much slower so I don't want to use this unless necessary

C-Loftus:
Closing this since other was merged

PR #87, opened 2024-07-10
Adds helpers for working with prompts

jaresty:
- Sometimes it's useful to set the prompt to your clipboard or view in the browser

### DIFF
--- a/GPT/gpt.talon
+++ b/GPT/gpt.talon
@@ -10,23 +10,23 @@ model <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
     result = user.gpt_apply_prompt(modelPrompt, text)
     user.gpt_insert_response(result, modelDestination or "")
 
-# Modifies a model command to be inserted as a snippet instead of a standard paste
+# Select the last GPT response so you can edit it further
+model take response: user.gpt_select_last()
+
+# Modifies a model command to be inserted as a snippet for VSCode instead of a standard paste
 # Otherwise same grammar as standard `model` command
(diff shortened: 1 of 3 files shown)


GPT/gpt.talon
```diff
@@ -10,8 +10,8 @@ model <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
     result = user.gpt_apply_prompt(modelPrompt, text)
     user.gpt_insert_response(result, modelDestination or "")
 
-# Select the last GPT response so you can edit it further
-model take response: user.gpt_select_last()
```
C-Loftus, line 14: We are getting rid of this? This doesn't seem contradictory to the new addition

jaresty: Oops. That looks like a mistake. I'm afk but will put that back if you don't beat me to it.

jaresty: Reverted that accidental deletion

C-Loftus:
Ok  I am fine with this now. I put this in gpt-beta since I am thinking this might be condensed into the grammar at some point but not sure what that will exactly look like now. I am using `model pass <prompt>` so it fits the syntax better. `model prompt` is a bit confusing imo. `model pass` signifies we are doing a noop where the source is the prompt itself.

Added some comments. Let's try to always get comments on the talon files whenever we add something.

C-Loftus:
Also moves the code that adds the snip modifier prompt _before_ the code that does the passing. This doesn't change any behavior atm, but in general as a best practice we should be checking special cases _after_ the prompt is finalized by checking modifiers.

jaresty:
Looks good to me!

PR #88, opened 2024-07-10
Add a destination to append to the clipboard

jaresty:
This is very useful in trying to chain prompts.


C-Loftus:
Seems fine to me. Changed it to be camel case one word instead of two words. Follows the pattern a bit better. 

Not sure how much usage this will get but it follows the grammar so happy to merge

C-Loftus (approved):

PR #89, opened 2024-07-11
Implement prompt modifiers

jaresty:
- 'snip' modifies both the prompt and the insertion method
- 'chain' automatically selects the inserted text after so that you can chain prompts together

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,8 +6,8 @@ Thank you for your interest in contributing to `talon-ai-tools`. We are happy to
 
 - For all PRs please describe the purpose of the PR and give an example of any non-trivial functionality.
 - Do not use external Python dependencies with `pip install`
-  - This pollutes the Talon Python interpreter and can have unintended side effects 
-- Prefix all actions exposed from a module with `gpt_` or another appropriate prefix to distinguish it in the global namespace 
+  - This pollutes the Talon Python interpreter and can have unintended side effects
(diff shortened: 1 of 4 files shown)


C-Loftus:
Made a couple stylistic edits. We don't want to expose actions globally in Talon if we don't need to. And we want to add a few doc strings above any .talon comand. Otherwise this should be all set now. Have not tested it since just got a new OS set up. Should be mainly fine though. 

Thanks for work on this!

Added a contribution.md file to help guide future contributors. Take a look over it and feel free to tweak wording.

C-Loftus:
What did you add? The latest commit added another copy of snip and removed the talon comment. I am fine with both `snip` and `chain`. Just make sure it is all documented and we don't have anything duplicated. Once you are good, feel free to merge.

jaresty:
It's not an exact copy-it doesn't include the prompt. It allows you to apply 'snippet' treatment to your clipboard when you paste using 'model snip clip' which is pretty useful. An alternative that would fit in the existing grammar would be something like 'model snip pass clip', but that didn't seem to work when I tried it. Since we have agreement on the first part of this I will merge what you approved and make a second PR for this change alone.

C-Loftus:
Ok. I think we might need to debug the `pass` command. I feel like there was some other issue with it. Would prefer to use pass until there is a great need.

jaresty:
Ah, I think 'pass' doesn't send the prompt to ChatGPT, which probably explains it.

PR #91, opened 2024-07-12
Provided way to apply snippets to clipboard alone

jaresty:
- Sometimes you want to paste the clipboard as a snippet so that you can easily modify it where it is appropriate
- In that case you're not really needing to use a prompt
- I tried using 'model snip pass' but it didn't work, so I introduced a new command the just passes a blank prompt which did the trick

I'm open to alternatives for this too.

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -159,7 +159,7 @@ class UserActions:
             text_to_process = prompt.removeprefix("ask")
             prompt = """Generate text that satisfies the question or request given in the input."""
         # If the user is just moving the source to the destination, we don't need to apply a query
-        elif prompt == "pass":
+        elif prompt == "pass" and modifier != "snip":
             return text_to_process
 
         match modifier:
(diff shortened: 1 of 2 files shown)


jaresty:
When you're using model snip you still want to pass the clipboard through ChatGPT.

jaresty:
If you don't pass the text through ChatGPT then none of the placeholders are added, so it's equivalent to 'model pass to clip' (i.e. 'snip' does nothing)

jaresty:
It is very useful to be able to add placeholders to clipboard contents, though, so I'd really like to have some way to support that.  I do think that 'snip pass' makes sense to use this way since 'snip' modifies the ChatGPT integration, which means you actually want to use ChatGPT even though you are using 'pass'.

C-Loftus:
If you want to have it return to the clipboard you can just use the destination though. `model snip please generate a dataclass to clip` should be fine? Pass is just a noop.

jaresty:
What I'm trying to do is not to return to the clipboard but to use the clipboard as input.  I.e. I've copied this loop (etc) and I want to paste it with placeholders.

jaresty:
I'm also open to defining a new command for this, i.e. 'model snip paste'

jaresty:
I found a satisfactory work around after all: `model snip answer clip` does just what I wanted.

PR #92, opened 2024-07-13
Add a command to run model apply clip on a cursorless target

jaresty:


C-Loftus:
Fine w me. You can merge.  I think some PRs are approaching a bit too specific to your use case if we continue adding a lot of stuff that does lots of permutations from selected text and the clipboard.  So this PR is fine but just as a heads up for future PRs if you are thinking about more stuff that that adds an significant complexity, perhaps make an issue first just to make sure we are on the same page.

jaresty:
Thanks for the heads up. With respect to this pull request-I'm also okay if you don't want to accept it. I will sometimes open pull requests as a way of demonstrating an idea that is difficult for me to explain in a issue. I'll leave this one open for a little bit to get some feedback from @pokey and also feel free to give your own feedback if you have any. Thank you 🙏

jaresty:
With respect to this grammar in particular-I was struggling with whether to make the cursorless target prefixed with 'to' since it felt a built awkward to say. However,.in this grammar it behaves more like a target than a destination so I was inclined to make the grammar agree.

C-Loftus:
For sure. I am fine with this. I was just moreso remarking about adding lots of permutations in general. I think this command is fine. You are free to merge.

jaresty:
Yeah I'm conscious of that too. I've appreciated the push back you gave on the 'snip' grammar and a quite satisfied with how it turned out. I think the main outstanding issue that's bugging me is that you can't interop between model targets and cursorless ones. Otherwise I'm pretty happy with how it's all working at present. I'm also motivated to keep it simple since it's better if more people can use it effectively.

C-Loftus:
@pokey Beyond the specifics of this PPR, do you have opinions on cursorless targets within our grammar more generally?

I know in the last meetup you said you didn't think they are a good idea. And I definitely get your point given the fact they are so complex and add to DFA compile time. Sometimes they aren't need and the same task can be accomplished by chaining two more verbose commands. 

That being said I do agree with @jaresty that without cursorless targets we end up needing to add helpers that break out of the grammar in sometimes awkward ways and add unnecessary verbosity. I think some complicated targets it adds cognitive load but for simple ones I think it is fine (ie refactoring an entire block)

I do generally like operating directly on cursorless targets. It reduces unnecessary chaining. 

I certainly wouldn't add more than one cursorless target to begin (ie so only a target probably, not both a target and a destination, but let me know if you think one is preferred)
But do you have any opinions on this?

PR #93, opened 2024-07-15
Add-image-source

jaresty:
- apparently the gpt-4o model supports images out of the box

https://platform.openai.com/docs/guides/vision

### DIFF
--- a/GPT/gpt.talon
+++ b/GPT/gpt.talon
@@ -20,6 +20,11 @@ model snip <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
     result = user.gpt_apply_prompt(modelPrompt, text, "snip")
     user.gpt_insert_response(result, modelDestination or "", "snip")
 
+# Applies a prompt to images in the clipboard
+model <user.modelPrompt> image [{user.modelDestination}]:
+    result = user.gpt_apply_prompt(modelPrompt, "", "image")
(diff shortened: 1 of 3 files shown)


jaresty:
I considered splitting up the prompts since many prompts don't apply to text and images both, however sometimes you take screenshots of things that include text and images. I think that rather than trying to artificially split them it may be perfectly reasonable to leave them altogether and rely on users to make a distinction about which prompt they want to use.

jaresty:
This might be a better approach: https://github.com/C-Loftus/talon-ai-tools/pull/94

GPT/gpt.talon
```diff
@@ -20,6 +20,11 @@ model snip <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
     result = user.gpt_apply_prompt(modelPrompt, text, "snip")
     user.gpt_insert_response(result, modelDestination or "", "snip")
 
+# Applies a prompt to images in the clipboard
+model <user.modelPrompt> image [{user.modelDestination}]:
```
C-Loftus, line 24: I like the syntax :)

C-Loftus:
Started to review this but this is superceded by the  newest open PR right?

jaresty:
Yes - I think the newest PR is the best of the options (IMO).  I left this here as an alternative idea.

PR #94, opened 2024-07-15
Make it so you can run 'model <prompt> clip' with images or text

jaresty:
If you have an image in your clipboard, this will try to use your prompt to process that image.


jaresty:
Going to make another change to this - making a draft for the moment.

PR #97, opened 2024-07-15
Add support for clipboard processing from images

jaresty:
- This add support for image processing
- requires the use of gpt-4o


jaresty:
I think this approach is workable if you are open to it. Because we are sending the requests much later than we are processing sources, I needed to make a signal that would allow me to know that I needed to pull from the clipboard later when sending the request to ChatGPT. This was the smallest change that I could find that seems to be workable-another alternative would be to build the request parts earlier where we are doing the source processing, but some of the other parts of the code rely on the text being raw so I wasn't sure that was a good idea for now.

jaresty:
This change allows you to use almost any screenshot or clip work contents and use them as input to every command. The possibilities are pretty open, and inspiring. You can use model blend to blend screenshots into text, or take a screenshot of a presentation and then use model snip to insert it into VSCode. The grammar is no different than it was before-the only change is that it check if the contents are an image to determine what to do.

lib/modelHelpers.py
```diff
@@ -53,22 +53,33 @@ def generate_payload(
...
                 "content": settings.get("user.model_system_prompt")
                 + additional_context,
             },
-            {"role": "user", "content": f"{prompt}:\n{content}"},
+            {"role": "user", "content": [{"type": "text", "text": prompt}, message]},
         ],
         "max_tokens": 2024,
         "temperature": settings.get("user.model_temperature"),
         "n": 1,
-        "stop": None,
```
jaresty, line 68: This was causing an exception when running with an image prompt but not with text. I didn't notice any behavioral difference however so I removed it.

jaresty:
This works really well. I just tried "model snip format html clip" with a prompt to generate html and a screenshot of a design I needed to make and it generated a snippet based on the visual design dynamically. This is pretty amazing stuff 😮

jaresty:
Incidentally appears that our current vision implementation is broken because the ChatGPT model is deprecated, however I didn't want to complicate this discussion by making a change there as well: https://github.com/tldraw/make-real-starter/issues/30

jaresty:
gpt-4o-mini launched today with vision and is cheaper than 3.5 turbo, apparently: https://openai.com/index/gpt-4o-mini-advancing-cost-efficient-intelligence/

GPT/gpt.py
```diff
@@ -244,7 +246,7 @@ def gpt_get_source_text(spoken_text: str) -> str:
         """Get the source text that is will have the prompt applied to it"""
         match spoken_text:
             case "clipboard":
-                return clip.text()
+                return "clip"
```
C-Loftus, line 249: Can we accomplish this PR without needing to change this to return a string which then is matched in the other function? Is it possible for us to return the actual source text here? 

If this is changed then `gpt_get_source_text`  doesn't return the source text for the clipboard anymore so it doesn't semantically match what it is doing. (i.e. if we wanted to call it elsewhere)

C-Loftus:
Generally looks fine w me if we can find a way to clean up that one comment I made. I am fine switching to gpt-4o-mini. Seems like a good choice.

For some reason I have felt gpt 3.5 turbo is not as good as it once was. Maybe it is just me but it seems the newer versions of the model are extremely focused on trying to return conversation-style responses, even with proper prompting

lib/modelHelpers.py
C-Loftus: Good w me if it works with you

GPT/gpt.py
jaresty: It's difficult to do this without a pretty significant change. The nice thing about this is that it transparently falls back if it's not an image in the clipboard, but we have to change the structure of the message we send to ChatGPT. We need some way to indicate that that should happen. This was the smallest change that I could think of to make this happen.

jaresty: Let's chat about this tomorrow-this syntax / strategy unlocks a lot of interesting possibilities that I'd love to share with you.  A bit hard to explain in text.

jaresty: I found a workaround that only makes the clipboard image a special case.

C-Loftus:
Ok looks good to me. Thanks. This is once again pretty clean. Nice work!

If you want to get the string, I still think we could b64 encode the image and pass it, but if you have other ideas don't want to break that. This is working well.

C-Loftus:
Oh and I switched to gpt-4o-mini. Don't see any reason not to use that. Gives users more support and has better performance. I don't think price is a bottleneck for anyone with this repo.

jaresty:
I debated returning the base sixty four encoded images as well, but you have to add a different type and pass the image URL when sending the message to ChatGPT, so you'd have to do some kind of a match to determine if it was a base sixty four encoded image, which is less clean than doing a direct string comparison I think.

C-Loftus:
Good to merge this now whenever you want. Change the magic string to be __IMAGE__ so it is essentially not possible to trigger accidentally.

PR #98, opened 2024-07-19
Add context to model requests for MacOS using a11y APIs

C-Loftus:
If the user is in VSCode and editing code, this code retrieves the entire context of the document without needing to manually copy and paste a second time. It then passes the entire ctx into the system prompt. gpt-4o-mini has a ctx window of 128k tokens (~95k words) which is huge and can easily take in the entire document. 

This dramatically improves responses and makes this much better.

To support this, in VSCode you need to enable `Toggle Screen Reader Accessibility Mode`. This is needed for us to use a11y apis with vscode. 

Since we are passing in a lot more tokens, it is good thing that the user needs to manually enable this feature. My only apprehension is that this would increase costs for users. But it is opt in so I think it is ok.


jaresty:
This looks really cool! I'll give it a whirl.

C-Loftus:
`max_tokens` is actually the max **output** tokens so I think 2048 is fine. I don't think we want the model able to output more than 2048 in most scenarios. GPT-4 supports 4k output tokens at the maximum. The issue is that if we allow large amounts of tokens then the model sometimes starts to fill up the token amount with unnecessary info. There are some scenarios in which this might be changed, but for right now, to keep things  relatively stable, I'd prefer to keep this as it is and then we can change things as necessary or add a max output token setting.

jaresty:
One thought I had in this one-I think it's a very interesting idea to include accessibility information generally. In code, in particular, I think the context necessary to provide better responses might not be in the current file, which is a limitation of this approach for code. I wonder if there's some way to "compress" the structure (I.e. include symbols and file paths or something) that would be better for code in particular. I still think this is useful, but wanted to share that thought in case you had ideas.

C-Loftus:
Going to merge this for now. Yes I agree it would be possible to pass in additional information, although more heuristics would have to be applied in order to make sure we aren't passing in excessive amounts of information. We can see how this does and iterate from there

PR #99, opened 2024-07-19
Adds various prompts for working with ChatGPT

jaresty:


C-Loftus:
Thanks! Added some prompts. Generally did not include anything that was project manager related or felt too specific. Changed a few or the prompt names. But these are pretty cool. TY!

PR #100, opened 2024-07-20
Adds a helper to search with a search engine

jaresty:
- Have you ever had to think about what you wanted to search for when you had the text of an error right in front of you?
- This allows ChatGPT to aid in that search, which can be surprisingly helpful and can also process images


C-Loftus:
I think I would prefer to keep this as a customization in your personal repo, just since it has a hard dependency on knausj with the search engines talon list and it doesn't require any changes to the lib code. 

I think I would like to focus on docs before I merge new features.

jaresty:
Nvm, figured it out: https://github.com/jaresty/mac-dotfiles/commit/7a407a5bf2235db009ad20ff144cda56102d6fb5

PR #101, opened 2024-07-20
Provide a hook for users to send their own context

jaresty:
- Users may have their own tools installed that they would like to use
- This hook should allow us to experiment more freely

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -113,6 +113,10 @@ class UserActions:
         GPTState.text_to_confirm = model_output
         confirmation_gui.show()
 
+    def additional_user_context():
+        """This is an override function that can be used to add additional context to the prompt"""
+        return []
+
     def close_model_confirmation_gui():
         """Close the model output without pasting it"""
         GPTState.text_to_confirm = ""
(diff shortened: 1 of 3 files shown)


jaresty:
The difference is that this is dynamic and can be overridden contextually. For example, one could implement the PR to include A11y data on macOS using this hook without having to change talon-ai-tools, since you have full access to the talon apis etc when it is called.

jaresty:
Some ideas I've had for this: include the current web site you are on in a browser, or run some custom git commands, etc.

GPT/gpt.py
```diff
@@ -113,6 +113,10 @@ def add_to_confirmation_gui(model_output: str):
         GPTState.text_to_confirm = model_output
         confirmation_gui.show()
 
+    def additional_user_context():
```
C-Loftus, line 116: Since we are already adding an additional context variable, let's change this to contextual user context or something similar perhaps?

C-Loftus (approved):
Looks good, you could change the name perhaps but might not be needed. Feel free to merge when you are happy. 

Is it better to use multiple items in the system prompt content array? I previously just had one big string. Unclear if one is better or not.

PR #106, opened 2024-07-29
Adds model context

jaresty:
- Model context allows you to accumulate user context that will be passed with every request
- You can automatically build up context using model thread
- You can also summarize the context to reduce the amount of context that needs to be sent


C-Loftus:
This is failing unfortunately since I can't figure out how to get the docs working. Info here for context: https://github.com/withastro/starlight/discussions/2158

jaresty:
Should we revert the James that started failing while we debug?

C-Loftus:
Would prefer to get that figured out first if possible so we can document this before merging. It is almost there just can't figure out the deployment path for github pages vs local.

jaresty:
Sounds good.

GPT/gpt.py
```diff
@@ -113,6 +115,26 @@ def add_to_confirmation_gui(model_output: str):
...
+        """Optimize the reused context to save tokens"""
+        optimize_context()
+
+    def gpt_optimize_thread():
+        """Optimize the thread to save tokens"""
+        optimize_thread()
+
+    def gpt_push_context(context: str):
+        """Add the selected text to the stored context"""
+        push_context(context)
```
C-Loftus, line 136: Might not be necessary to extract the context helper functions as a separate function if we aren't doing anything but call that function. But perhaps I am missing the background.

GPT/gpt.talon
```diff
@@ -36,6 +36,19 @@ model apply [from] clip$:
...
+    user.gpt_push_context(text)
+model context clear: user.gpt_clear_context()
+model context optimize: user.gpt_optimize_context()
+
+model thread <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
+    text = user.gpt_get_source_text(modelSource or "")
+    result = user.gpt_apply_prompt(modelPrompt, text, "thread")
+    user.gpt_insert_response(result, modelDestination or "", "thread")
+model thread new: user.gpt_new_thread()
+model thread optimize: user.gpt_optimize_thread()
```
C-Loftus, line 50: I am not very clear on how thread is different from normal context. So this puts the previous response in the context right?

I feel that for users this might be confusing. I wonder if we can use a different term.

lib/modelHelpers.py
```diff
@@ -42,17 +112,36 @@ def make_prompt_from_editor_ctx(ctx: str):
...
+    prompt: str, content: str, tools: Optional[list[Tool]] = None, modifier: str = ""
 ) -> Tuple[Headers, Data]:
     """Generate the headers and data for the OpenAI API GPT request.
     Does not return the URL given the fact not all openai-compatible endpoints support new features like tools
     """
-    notify("GPT Task Started")
-
+    global stored_context
+    global thread_context
+    notification = "GPT Task"
```
C-Loftus, line 137: This is no longer going to say GPT task started I believe for normal requests

C-Loftus (commented):
I think this is good. I think we might want to alter some of the terminology just since thread is a bit confusing in my opinion since we have chain which is a bit similar. I wonder if we can combine some of the semantics there but obviously keep the distinct behavior. 

In general the code seems fine. Just seems like there might be some duplication we could condense. 

I think that in the future we can probably create a capture for context that accepts either a cursorless target, the clipboard or the a11y context.

GPT/gpt.talon
jaresty: Thread and context are separate-context will be sent with every request regardless. Thread conversations will continue only if you use the thread command. This lets you set up the some baseline context that you can use for all conversations (think custom gpt) that you curate. That context is valuable since you spent the time to pick it out specifically. The thread content is specific to one conversation and you probably want to get rid of it when you start the next topic.

GPT/gpt.py
jaresty: I did that to isolate the global usage, mostly. It seemed cleaner to me.

jaresty:
Some other alternatives to thread:
- chat
- continue
- converse
- memory
- remember

lib/modelHelpers.py
jaresty: I restored the original wording.

GPT/gpt.talon
C-Loftus: This is cool and makes sense. I think it is a good idea. I think we need to think about ways to manage state with this however. I think it is going to be confusing for a user to know what is in the thread or how to use these.

I think thread is fine as a keyword. It is shorter than others I can think of. I was thinking of threading in software not threads on a chat app. 

I am thinking we might want to have some sort sort of visual indicator that shows the number of lines / first few words in the stored context or something analogous. Just so it is very clear what is being sent/stored. And then we can use the same indicator when a request is in process or something . Does that seem reasonable?

jaresty: I like that idea but I'm not sure how to implement it. Can you contribute that functionality? If it seems challenging, we could also plan to to it as a follow up. This is really useful as-is, and the confusion isn't too bad in my experience since it's quite obvious from the responses you get that there is more in the context than you thought. I've been using this branch all week.

jaresty: As an mvp, we could make "thread" and "context" into sources that can be "passed"...?

jaresty: Now that I'm thinking of it, context might make sense as a destination too, as an alternative to the custom grammar "model context add", you could "model pass to context". Though you'd still need something custom to "clear" it. I'm not sure if that's better.

jaresty: I played with this a little bit and it's not very straightforward to fit it into the existing grammar because I'm not sure how to change the code to fit it in. I'm still finding it quite useful as is.

jaresty: I think I found a middle ground that adopts some grammar from the existing pass command.

C-Loftus: Yes do not worry about an indicator right now. That is some thing I would like to add eventually but we don't need to do now. 

For this because it is not really doing a model operation to my understanding but only getting a model operation ready, I wouldn't worry too much about fitting it into the grammar. Do whatever is best for you

C-Loftus:
New changes look good to me

jaresty:
I might take one more pass at making the context into a source / destination since I think it's doable and could be useful. I'll probably leave the thread grammar alone. Adding a11y to context sounds very useful-lets you skip "select all" and even the screenshot. That's a great idea, and it will work across most applications.

jaresty:
Ok, I think I'm all set.  Let me know if you still think this looks good.

C-Loftus:
Yup this all looks good to me. Might clean up a few minor things in terms of code style in a future PR related to pushing the a11y context. But generally looks good w me. Nice work!

Before you merge could you just add one line above each command you added in the `.talon` file. We can worry about more thorough documentation in the future once other behavior is solidified.

jaresty:
Added some spacing - if this looks good, please merge 🙏

C-Loftus:
Sorry when I said I line above I meant a line with comment info. Added that. Only thing I was thinking is that may be want to have thread be a destination / source as well. Can iterate on this later. Alternatively not having it be a source destination is more explicit for users looking into the code. So not sure.

Also  might want to use the word `summarize` instead of optimize. Since optimize is a bit vague regarding what optimization it is doing. We can think about this in a future PR if youd like.

jaresty:
I think it's pretty straightforward to add thread as a source.  Not sure about destination-what did you have in mind?

jaresty:
with respect to summarize, I was considering getting rid of it in favor of `model summarize context to new context`.  Could maybe do the same for thread.

C-Loftus:
> with respect to summarize, I was considering getting rid of it in favor of `model summarize context to new context`. Could maybe do the same for thread.

I like that!

> I think it's pretty straightforward to add thread as a source. Not sure about destination-what did you have in mind?

As a destination it would just add to the thread right?

jaresty:
Oh, yeah, I think that could work.  It's a bit confusing only b/c the normal behavior if you use `model thread` would be to append to the thread, so as a destination I think you'd not be using `model thread` to make that sensible to use.

jaresty:
I can make a PR for this later, but feel free to add it yourself too :)

PR #107, opened 2024-07-29
Add application context

jaresty:
- This context allows the LLM to understand context relevant questions
- For example, when you use 'model ask' in slack, it knows that you might be asking about slack messages. When you use it inside of Jira, it knows that you might be interested in jira tickets

### DIFF
--- a/lib/modelHelpers.py
+++ b/lib/modelHelpers.py
@@ -54,10 +54,14 @@ def generate_payload(
 
     language = actions.code.language()
     additional_context = (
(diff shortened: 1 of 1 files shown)


C-Loftus (approved):
Looks good to me but just for clarification is breaking the system prompt into multiple items better than one larger block? Wasn't sure either way but we can do whatever is optimal.

jaresty:
I think it helps the LLM understand that each of these is a different topic.

PR #114, opened 2024-07-31
Cursorless should only be applied to simple prompts

jaresty:
- cursorless is too similar to natural language
- therefore we should not use it with model commands that apply to natural language


C-Loftus:
It would be nice to have `model please` on a cursorless target and maybe we can do that in the future but I agree this is probably the right choice

C-Loftus (approved):

PR #116, opened 2024-08-02
Unify-thread-and-context-interface

jaresty:
this unifies the model context and model thread interfaces


C-Loftus:
Did a bit of cleanup on this. Didn't change any behavior. 

- I removed any usage of global variables and put any persistent state within the GPTState class. 
- I removed any functions that were exposed globally that weren't being called globally and were leftovers from previous factoring.
- I removed any parameters or functions that weren't necessary
- I changed it so we aren't passing magic _THREAD_ variables and the like. Since that normal text we don't need to pass get it later. 
- I made the process of constructing the messages a bit cleaner
- I improved the prompts a bit so it is very clear that the thread is previous conversation with the model and not just arbitrary context. 

One thing we need to be careful of is having a lot of special cases or things which require us to deviate from existing patterns. 

In the future we might want to investigate a way to make model thread less verbose so the user can have a more natural chat with the model. That might require a special talon mode to be toggle for natural speech or something analogous. But this is out of scope and not super necessary right now

jaresty:
Before, the context supported clipboard images as well, I think this broke that, since you are now appending context as strings.

jaresty:
I was relying on the context support of snapshots - I'd like to preserve that behavior b/c I don't always have a good way to select text and I use it to work with designs (like figma, etc)

jaresty:
Though maybe I misread.  I'll give it a try.

jaresty:
Yeah, the image support is broken.  Reproduction steps:
1. grab a screenshot
2. 'model pass clip to context'
3. Run a prompt

Here is what I get now:
```
Context: __IMAGE__ - The text discusses the integration of thread and context interfaces in C programming, aiming to unify them for better performance and simplicity. The proposal includes changes to the current API, potential benefits, and considerations for backward compatibility.
```
having passed this image  to my context.
image

jaresty:
'model summarize clip' seems broken too:

Unify the thread and context interface in the C project.

jaresty:
the unify information is coming from my window title, I believe

jaresty:
This is how 'model summarize' worked before with this image in my context:

Preserve context support of snapshots for selecting text, as it is useful for working with designs (e.g., Figma).

image

jaresty:
Here's how I was planning to use the ability to combine clipboard images with text in context in order to produce more useful responses:

1. I was going to copy CSS definitions from my website
2. use model thread to collect a list of tailwind classes and put that into my context
3. from figma you can copy CSS for an element and everything beneath it which provides context on exactly what font sizes and such are expected
4. I was going to add images of the elements that I needed to generate to my context
5. then I was going to use model prompts to build up the page

C-Loftus:
Ok I was not aware you wanted to support an image as context. You can either revert my most recent commit and add the changes to your last commit or make the changes to mine. However, we do need to clean it up a bit. I do not have bandwidth to do this unfortunately. 

`model summarize clip` still works for me with text but I agree no longer with images. That is because I changed each category in the prompt list to be one string instead of a list of messages. Perhaps that should be reverted.

jaresty:
since I'm not really sure how to integrate these changes without breaking the behavior I'm after at the moment, while I am okay with trying to pursue a better architecture long term, I think my preference would be to revert those changes for the moment and redo them when you have bandwidth-does that sound okay to you? I don't really have bandwidth to try and figure out how to integrate them myself at the moment-sorry!

C-Loftus:
Ok let's just keep this open then as a draft for the time being. I think it is going to be much easier to make a few changes to the latest commit here rather than merge and redo it again later. 

For what it is worth, I didn't change anything architecturally. I just made the code a bit cleaner since we are getting a lot of helper functions and it is becoming harder to follow the behavior. There are some style issues around global/unused variables/unnecessary helper functions/type errors that I don't feel comfortable merging since it makes maintaining/debugging things harder. 
Note that this isn't because of this commit or you, it is just the generally trend of this repo lately and I want to try and make this repo as easy to maintain as possible since I don't have a ton of bandwidth for it anymore with my new job

 You should be able to use this branch for your personal use. I won't be adding anything else upstream for the time being. 
I am unclear if the most recent revert commit you did actually reverted things completely. It seems like some of my changes are still present. I think we should hold off on changes before we make things harder for ourselves.

C-Loftus:
Thank you!

Issue #120, opened 2024-08-02
Refactor codebase with more flexible interfaces for a variety of modifiers


C-Loftus:
With the amount of modifiers we have, the current abstraction in the `.talon` file doesn't necessarily make sense.  We should probably only expose one function to talon and then do all internal logic in Python in a way that is more flexible. 

https://github.com/C-Loftus/talon-ai-tools/blob/1be3bbe938b6d85caa1542ffa7550adf62899e1c/GPT/gpt.talon#L8-L11

We shouldn't necessarily use these three functions since it requires us to awkwardly pass around arguments in places where they aren't semantically useful, simply for the purpose of passing them to the next stage in the pipeline. So we will keep the same source and destination pattern in talon but change the logic in Python.

We should also have some sort of generic interface for context and state that doesn't require use to manually check the type of the data at each step. This should probably be done with an ABC or protocol in python.

PR #122, opened 2024-08-03
Refactor-single-funk

jaresty:
This is an attempt to address the issues raised in #120

### DIFF
--- a/GPT/lists/modelModifier.talon-list
+++ b/GPT/lists/modelModifier.talon-list
@@ -1,0 +1,11 @@
+list: user.modelModifier
+-
+
+# Construct snippet with the response
+snip: snip
+
+# Keep track of the questions and responses as a conversation
+thread: thread
+
+# Select the response after insertion so you can apply subsequent prompts on it
+chain: chain
--- a/lib/modelHelpers.py
+++ b/lib/modelHelpers.py
@@ -39,10 +39,10 @@ def push_context(context: str):
     actions.app.notify("Appended user context")
 
 
-def push_thread(context: str):
+def push_thread(context: dict[str, any]):
     """Add the selected text to the current thread"""
     global thread_context
-    thread_context += [format_message(context)]
+    thread_context += [context]
     actions.app.notify("Appended to thread")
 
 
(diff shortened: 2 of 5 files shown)


lib/modelHelpers.py
```diff
@@ -32,40 +32,20 @@ def new_thread():
     actions.app.notify("Created a new thread")
 
 
-def push_context(context: str):
+def push_context(context: dict[str, any]):
     """Add the selected text to the stored context"""
     global stored_context
```
C-Loftus, line 37: Good with me in this PR, just as a comment I think we will eventually refactor this out into the GPTState class but no rush on this.

GPT/lists/modelModifier.talon-list
```diff
@@ -0,0 +1,11 @@
...
+-
+
+# Construct snippet with the response
+snip: snip
+
+# Keep track of the questions and responses as a conversation
+thread: thread
+
+# Select the response after insertion so you can apply subsequent prompts on it
+chain: chain
```
C-Loftus, line 11: I think this is a good idea. I think it adds clarity to how the model is used.

lib/modelHelpers.py
```diff
@@ -136,23 +116,31 @@ def make_prompt_from_editor_ctx(ctx: str):
 
 
 def format_message(content: str):
```
C-Loftus, line 118: Good for right now in this PR; I think we may eventually want to refactor this to use [Python protocols](https://peps.python.org/pep-0544/) to enforce a static structurally typed interface. But nothing wrong with this if it works for the time being. 

I can do that python class refactoring and experiment a bit since I know that it is very Python-specific. Will hold off any changes to this branch so I don't disrupt this PR; learned my lesson with the last PR and want to make things small and atomic.

lib/modelHelpers.py
```diff
@@ -32,40 +32,20 @@ def new_thread():
...
+def push_context(context: dict[str, any]):
     """Add the selected text to the stored context"""
     global stored_context
-    stored_context += [format_message(context)]
+    stored_context += [context]
     actions.app.notify("Appended user context")
 
 
-def push_thread(context: str):
+def push_thread(context: dict[str, any]):
```
C-Loftus, line 42: Once again not relevant to this PR but we will eventually want to have this have the proper roles. i.e. `assistant` for the model responses. That way our model is truly seeing the model history with who did what and not just a series of user strings.

C-Loftus:
Made some comments. Nothing wrong with this though, moreso just for alignment / context for future goals.

Once you are satisfied with this I can then look at refactoring with some of the Python-specific stuff so we have nice generic interfaces.

PR #125, opened 2024-08-04
Add-writing-prompts

jaresty:
Try this:
- use dictation mode to brainstorm a rough topic idea
- use model add questions to add questions to it and answer the questions one by one
- use model format outline to generate a rough outline for that
- use model format prose to take that outline and turn it into text that reads well

### DIFF
--- a/GPT/lists/staticPrompt.talon-list
+++ b/GPT/lists/staticPrompt.talon-list
@@ -48,3 +48,8 @@ generate code: The following plaintext describes a process in code in the langua
 update comments: Act as a software engineer. The following code may be missing comments or the comments could be out of date. Please update the comments. If you are unsure how to comment something, ask a question in a comment instead. Return just the code and not any explanations.
 clean code: Act as a software engineer. Reduce any duplication in the selected code and improve it to be more idiomatic and clear for other users. However, do not change the behavior or functionality. Return just the code and not any explanations.
 improve semantics: The following is an HTML document. Keep the same structure and layout but if it is needed, change any elements to use proper semantic HTML and make sure it is implementing best practices for user accessibility. Output just the HTML and not any extra explanations.
+
+## PROMPTS TO HELP WITH WRITING
+add questions: Help me explore this question from multiple perspectives. For each perspective, ask follow-up questions and indicate what perspective is being taken. Annotate the dialogue to show its a conversation between you and me, as this will be shared on Slack and needs to be clear that its a generated response. Respond with only the perspectives and the questions they generated.
+format outline: Create an outline that encapsulates the text below. Keep the number of sections between three and five to optimize for human working memory. Return just the outline.
+format prose: As an editor, format the following text as prose. You can have headings and paragraphs. Trying to avoid bullets. Reorder and add transitions as necessary to make the document flow. Return just the text.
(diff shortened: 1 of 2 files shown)


GPT/lists/staticPrompt.talon-list
```diff
@@ -48,3 +48,8 @@ generate code: The following plaintext describes a process in code in the langua
...
 clean code: Act as a software engineer. Reduce any duplication in the selected code and improve it to be more idiomatic and clear for other users. However, do not change the behavior or functionality. Return just the code and not any explanations.
 improve semantics: The following is an HTML document. Keep the same structure and layout but if it is needed, change any elements to use proper semantic HTML and make sure it is implementing best practices for user accessibility. Output just the HTML and not any extra explanations.
+
+## PROMPTS TO HELP WITH WRITING
+add questions: Help me explore this question from multiple perspectives. For each perspective, ask follow-up questions and indicate what perspective is being taken. Annotate the dialogue to show its a conversation between you and me, as this will be shared on Slack and needs to be clear that its a generated response. Respond with only the perspectives and the questions they generated.
```
C-Loftus, line 53:
```suggestion
add questions: Help me explore this question from multiple perspectives. For each perspective, ask follow-up questions and indicate what perspective is being taken. 
```
I think the part about having it annotated as a dialog for sharing with others might be too specific.

GPT/lists/staticPrompt.talon-list
```diff
@@ -48,3 +48,8 @@ generate code: The following plaintext describes a process in code in the langua
...
+format outline: Create an outline that encapsulates the text below. Keep the number of sections between three and five to optimize for human working memory. Return just the outline.
+format prose: As an editor, format the following text as prose. You can have headings and paragraphs. Trying to avoid bullets. Reorder and add transitions as necessary to make the document flow. Return just the text.
```
C-Loftus, line 55: When would this be used?

C-Loftus, line 55:
```suggestion
format prose: As an editor, format the following text as prose. You can have headings and paragraphs. Avoid using bullet points. Reorder and add transitions as necessary to make the document flow. Return just the text.
```

GPT/lists/staticPrompt.talon-list
```diff
@@ -48,3 +48,8 @@ generate code: The following plaintext describes a process in code in the langua
...
 clean code: Act as a software engineer. Reduce any duplication in the selected code and improve it to be more idiomatic and clear for other users. However, do not change the behavior or functionality. Return just the code and not any explanations.
 improve semantics: The following is an HTML document. Keep the same structure and layout but if it is needed, change any elements to use proper semantic HTML and make sure it is implementing best practices for user accessibility. Output just the HTML and not any extra explanations.
+
+## PROMPTS TO HELP WITH WRITING
+add questions: Help me explore this question from multiple perspectives. For each perspective, ask follow-up questions and indicate what perspective is being taken. Annotate the dialogue to show its a conversation between you and me, as this will be shared on Slack and needs to be clear that its a generated response. Respond with only the perspectives and the questions they generated.
+format outline: Create an outline that encapsulates the text below. Keep the number of sections between three and five to optimize for human working memory. Return just the outline.
```
C-Loftus, line 54: A bit similar to summarize but I think it is ok.

C-Loftus:
Fine with me. Made a few suggestions. I think there are a few ways in which we can make it more generalizable beyond our specific workflows. I think the pattern of adding questions, summarizing into an outline, and then condensing is a fair pattern.

jaresty:
Let me give the new prompts a whirl before we merge-I want to make sure they're still working well.

jaresty:
Okay I tested it out and it seems to be working still. Works for me!

C-Loftus (approved):

PR #126, opened 2024-08-04
Baseline refactoring to prepare for cleanups

C-Loftus:
Doesn't change any behavior, just prepares the codebase a bit by moving a few things to new files and beginning to add a new class for a generic GPTMessage. I need to experiment with how we will use this with images and context. 

Instead of passing things through, now that we have one larger function, we can just set a state variable at the start, check it right before the query, and then reset it at the end of the query. I think this might be easier and we don't need to pass things around. Will explore this later.


C-Loftus:
Merging since this doesn't affect any user-facing behavior

PR #127, opened 2024-08-04
Use a better state class; more general refactor work

C-Loftus:
Changes the code to use a central state class instead of global variables. This makes the code cleaner and easier to extend.

Any methods dealing with state like threading or contact we want to put within this class to separate from pure functions or talon user facing code.

Cleaned up a few other things. gpt.talon included "normal" as the default but "normal" was not being checked anywhere. So just used "" instead. 

Don't believe this causes any regressions and is just a refactor.

### DIFF
--- a/GPT/gpt-contextual.talon
+++ b/GPT/gpt-contextual.talon
@@ -1,0 +1,6 @@
+# Clear the context stored in the model
+model context clear: user.gpt_clear_context()
+
+# Create a new thread which is similar to a conversation with the model
+# A thread allows the model to access data from the previous queries in the same thread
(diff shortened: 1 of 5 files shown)


C-Loftus:
Should not regress any behavior. Once this is merged and reviewed, lets take a look at #128 and make sure our context / thread code is very explicit. Didn't want to touch that and cause a regression.

jaresty (approved):
🚢 I gave this a try and it seems to work great

C-Loftus:
Merging for now, should be fine from what I can tell as well.

Issue #128, opened 2024-08-04
Clarify some of the context behavior before refactoring


C-Loftus:
https://github.com/C-Loftus/talon-ai-tools/blob/69845f3fea97c6d7ca83e8df0d30bc7d51a50db9/lib/modelHelpers.py#L175-L185

There are a few places within the payload generation function where the behavior is a bit unclear. Here for instance we have a check against the variable `__CONTEXT__` but this literal does not appear anywhere in the code. Just want to be clear on some things before I refactor the context code.

We are also add the thread just by adding the list to the previous one. Wanted to confirm this is proper.

We should try to be using as strict type checking as possible to make our code very explicit. We have a lot of edge cases and every PR risks breaking things.

jaresty:
> We are also add the thread just by adding the list to the previous one. Wanted to confirm this is proper.


The thread behavior is intentional there-we don't want to include the thread on every request, just the requests that are using the thread modifier (at least until we decide to do some kind of a toggle)

jaresty:
> We should try to be using as strict type checking as possible to make our code very explicit. We have a lot of edge cases and every PR risks breaking things.

100% agree. I'm doing the best I can-unfortunately there are not types in talon script so those are a little challenging to catch.

C-Loftus:
> > We should try to be using as strict type checking as possible to make our code very explicit. We have a lot of edge cases and every PR risks breaking things.
> 
> 100% agree. I'm doing the best I can-unfortunately there are not types in talon script so those are a little challenging to catch.

Sounds good. And that above comment was not about your code, just a statement about the codebase in general and something I was considering adding to the contributing docs.

C-Loftus:
Fixed with latest PR #129

PR #129, opened 2024-08-04
This exposes a simple run prompt method

jaresty:
- some of my custom extensions were failing because there was no longer an API for running a prompt and getting a response in talon script

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -83,9 +83,9 @@ class UserActions:
        Do not output comments, backticks, or natural language explanations.
        Prioritize SQL queries that are database agnostic.
         """
-        return gpt_query(format_message(prompt), format_message(text_to_process))[
-            "text"
-        ]
+        return gpt_query(format_message(prompt), format_message(text_to_process)).get(
+            "text", ""
+        )
 
     def gpt_clear_context():
         """Reset the stored context"""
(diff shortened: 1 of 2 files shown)


GPT/gpt.py
```diff
@@ -133,8 +133,16 @@ def gpt_select_last():
         for _ in lines[0]:
             actions.edit.extend_left()
 
-    def gpt_apply_prompt(mode: str, prompt: str, source: str, destination: str):
+    def gpt_apply_prompt(
+        mode: str, prompt: str, source: str, destination: str
+    ) -> dict[str, str]:
         """Apply an arbitrary prompt to arbitrary text"""
+        response = actions.user.gpt_run_prompt(mode, prompt, source)
+        actions.user.gpt_insert_response(format_message(response), destination, mode)
```
jaresty, line 141: notably we are unwrapping and rewrapping the message here. I went down the path of trying to skip doing that and it tended up being a much bigger change, so I think this is better for now.

jaresty:
I also addressed #128

C-Loftus (approved):
Looks good to me. Thank you. My understanding of this PR,  correct me if I am wrong, is that we need to return the value into talon so then a user can run an arbitrary function of it if they would like.

I think a raw modifier may be interesting at some point, but not needed at the moment.

Minor merge conflict but feel free to merge once you are happy.

GPT/gpt.py
C-Loftus: Fine with me. My plan is to eventually have a GPTMessage class that makes this a bit easier to reason about.

jaresty:
That's exactly right-some of my custom scripts and I'm sure others as well were relying on the return value of the old function. This is a new api but at least it still works for that!

PR #130, opened 2024-08-05
Add a new destination for an imgui window and fix tts / cursorless destination

C-Loftus:
I believe the TTS and cursorless pasting needed to get the message unpacked. 

Added imgui support by using `to window` as the destination name since it is very easy to add and nicer than the browser in many ways


jaresty (approved):
Makes sense to me!

PR #131, opened 2024-08-05
Simplify format handling of insertion

jaresty:
- inserting text does not really care about formats
- persisting thread and context requires formats only on inputs coming directly from get source (due to clipboard images)
- when inserting responses from an llm those are always text (for now) so that's better to isolate there


jaresty:
If we ever start handling non textual responses from the llm, we may want to consider some other mechanism for differentiating how to store context (textual versus structured)-for the moment though, it seems to me that most of our insertion strategies are relying on text.

PR #132, opened 2024-08-05
Adjust verbosity of model window commands

jaresty:
- if you have the model window open, I think that the "model" is implied
- I think it's also useful to add chain to these commands
- This also aligns the naming of the model output from 'model take response' (I think we should use the same name everywhere for the model output/response)

### DIFF
--- a/lib/modelConfirmationGUI.py
+++ b/lib/modelConfirmationGUI.py
@@ -24,6 +24,14 @@ def confirmation_gui(gui: imgui.GUI):
         actions.user.gpt_select_last()
 
     gui.spacer()
+    if gui.button("Pass response to context"):
+        actions.user.pass_context_model_confirmation_gui()
+
+    gui.spacer()
+    if gui.button("Pass response to thread"):
+        actions.user.pass_thread_model_confirmation_gui()
+
+    gui.spacer()
     if gui.button("Copy response"):
         actions.user.copy_model_confirmation_gui()
 
@@ -45,6 +53,18 @@ class UserActions:
         GPTState.text_to_confirm = ""
         confirmation_gui.hide()
         ctx.tags = []
+
+    def pass_context_model_confirmation_gui():
+        """Add the model output to the context"""
+        actions.user.gpt_push_context(GPTState.text_to_confirm)
+        GPTState.text_to_confirm = ""
+        actions.user.close_model_confirmation_gui()
+
+    def pass_thread_model_confirmation_gui():
+        """Add the model output to the thread"""
+        actions.user.gpt_push_thread(GPTState.text_to_confirm)
+        GPTState.text_to_confirm = ""
+        actions.user.close_model_confirmation_gui()
 
     def copy_model_confirmation_gui():
         """Copy the model output to the clipboard"""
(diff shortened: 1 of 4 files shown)


lib/modelConfirmationGUI.py
```diff
@@ -54,5 +59,12 @@ def paste_model_confirmation_gui():
             return
         else:
             actions.user.paste(GPTState.text_to_confirm)
+            GPTState.last_response = GPTState.text_to_confirm
+            GPTState.last_was_pasted = True
             GPTState.text_to_confirm = ""
             confirmation_gui.hide()
+
+    def chain_model_confirmation_gui():
```
C-Loftus, line 67: If we are just calling two functions that are already globally exposed I don't think we need a new function. We can just call those directly in the .talon file. That makes it more explicit.

C-Loftus:
The reason I have this more verbose is given the fact that it could conflict with other user commands related to pasting since it is pretty generic. 

But I think this is fine, and probably not too likely to conflict. Ok with merging if you want.

jaresty:
That's a good point-I followed the pattern of the homophones window to enable it only when the window is open now.

lib/modelConfirmationGUI.py
```diff
@@ -14,45 +15,54 @@ def confirmation_gui(gui: imgui.GUI):
...
     def close_model_confirmation_gui():
-        """Close the model output without pasting it"""
+        """Close the model output without pasting itp"""
         GPTState.text_to_confirm = ""
         confirmation_gui.hide()
+        ctx.tags = []
 
     def copy_model_confirmation_gui():
-        """Copy the model output to the clipboard"""
+        """Copy the model ctx.tags = ["user.model_window_open"] to the clipboard"""
```
C-Loftus, line 50:
```suggestion
        """Copy the model output to the clipboard"""
```

lib/modelConfirmationGUI.py
```diff
@@ -14,45 +15,54 @@ def confirmation_gui(gui: imgui.GUI):
...
 class UserActions:
     def add_to_confirmation_gui(model_output: str):
         """Add text to the confirmation gui"""
+        ctx.tags = ["user.model_window_open"]
         GPTState.text_to_confirm = model_output
         confirmation_gui.show()
 
     def close_model_confirmation_gui():
-        """Close the model output without pasting it"""
+        """Close the model output without pasting itp"""
```
C-Loftus, line 44:
```suggestion
        """Close the model output without pasting it"""
```

jaresty:
I'd like to also add context/thread to this.

C-Loftus:
Looks fine but a bit worried that 6 buttons on one window is a bit much. Let's keep it for now but might want to refactor this later or remove buttons.

jaresty:
It is a lot of buttons so I definitely would support a way to condense it if it's possible-however I'm pretty glad that there is some way to interactively select how you want to handle the model response. It's pretty powerful!

C-Loftus:
Talon currently can't do horizontal buttons or columns in imgui. So I guess for right now we can just merge this since there isn't a better option. I am not sure if model chain is needed since it is just a helper and doesn't add any behavior really, but we can keep it

jaresty:
Personally I really like model chain-I think it's really handy to have it in the window UI especially since you're pausing for a moment to think about the response you're more likely to know that you want to select it when you input it as well. I'm gonna merge this for now and we can think about it some more-thanks!

PR #133, opened 2024-08-05
Make the system prompt stricter 

C-Loftus:
We don't want natural language explanations to be output when generating code.

### DIFF
--- a/lib/talonSettings.py
+++ b/lib/talonSettings.py
@@ -52,9 +52,10 @@ mod.setting(
 mod.setting(
     "model_system_prompt",
     type=str,
-    default="You are an assistant helping an office worker to be more productive. Output just the response to the request and no additional content. Do not generate any markdown formatting such as backticks for programming languages unless it is explicitly requested.",
+    default="You are an assistant helping an office worker to be more productive. Output just the response to the request and no additional content. Do not generate any markdown formatting such as backticks for programming languages unless it is explicitly requested. If the user requests code generation, output just code and not additional natural language explanation.",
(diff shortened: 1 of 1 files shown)


lib/talonSettings.py
```diff
@@ -52,10 +52,11 @@ def modelSimplePrompt(matched_prompt) -> str:
 mod.setting(
     "model_system_prompt",
     type=str,
-    default="You are an assistant helping an office worker to be more productive. Output just the response to the request and no additional content. Do not generate any markdown formatting such as backticks for programming languages unless it is explicitly requested.",
+    default="You are an assistant helping an office worker to be more productive. Output just the response to the request and no additional content. Do not generate any markdown formatting such as backticks for programming languages unless it is explicitly requested. If the user requests code generation, output just code and not additional natural language explanation.",
```
jaresty, line 55: This was the suggestion from ChatGPT:

You are an assistant helping an office worker to be more productive. Output only the response to the request with positive language. Generate markdown formatting only if explicitly requested. When the user requests code generation, provide only the code.

C-Loftus: I think that is probably fine as well. It is slightly less verbose.  If you try that and get better responses feel free to merge that instead. If you test it and prefer that I'm personally fine with either way as long as it solves the issues of the model adding unnecessary natural language comments to code generation. 

I am a bit skeptical of the models ability to make meta suggestions about how to prompt itself.

jaresty (approved):
Let's merge this for now and I will try mine out with my local configuration

PR #135, opened 2024-08-06
Move the snip handling to the prompt generation

jaresty:
- we're handling other modifiers here as well, so this makes more sense I think


C-Loftus:
I thought we wanted to use an array for the system prompts? I feel like we don't really want this change since it switches to using a big string instead of the array of items like we had before.

jaresty:
I think we do want to use an array, yeah - I think what we had wasn't actually doing that though.  I can look at that later.

C-Loftus (approved):

PR #136, opened 2024-08-06
Model is a fine name, but it is two syllables

jaresty:
- some people may want to rename their model to avoid conflicts with their particular configuration it or to make it easier for them to say the name

### DIFF
--- a/.docs/src/content/docs/guides/customizing.mdx
+++ b/.docs/src/content/docs/guides/customizing.mdx
@@ -14,3 +14,42 @@ import FetchedData from "../../../components/FetchedData.astro";
 You do not need to fork the repository to add your own custom prompts. Copy the file below, place it anywhere inside your talon `user/` directory and follow the pattern of the key value mapping.
 
 <FetchedData data="https://raw.githubusercontent.com/C-Loftus/talon-ai-tools/main/GPT/lists/customPrompt.talon-list.example" />
+
+
+
+## Advanced Customization
+
+### Configuring Model Name
+
+The model name is used to invoke the large language model. By default it is named 'model' and you can use 'model' when you invoke any commands. For example you can ask the model to do something for you by saying 'model ask \<text\>'. You can override this in your own configuration without changing this repository. To do so just create another talon list with the same name and a higher specificity. Here is an example that you can copy and past into your own configuration files
+
+```yml
+list: user.modelAlias
+-
+
+# The default model name
+model: model
+```
+
+
+### Providing Custom User Context
+
+In case you want to provide additional context to the LLM, there is a hook that you can override in your own python code and anything that is returned will be sent with every request. Here is an example:
+
+```py
+from talon import Context, Module, actions
+
+mod = Module()
+
+ctx = Context()
+
+
+@ctx.action_class("user")
+class UserActions:
+    def contextual_user_context():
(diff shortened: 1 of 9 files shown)


GPT/lists/modelName.talon-list
```diff
@@ -0,0 +1,5 @@
+list: user.modelName
+-
+
+# The default model name
+model: model
```
jaresty, line 5: aegis recommended using a talon list for this even though it is a single word.

C-Loftus:
This is OK but it hurts readability for the commands. Now that we have the modifier as well I think it is confusing for new users. 

New users can't look at the talon files anymore to get a sense of what the commands do. 

Model name as a list is a bit confusing since it implies we could specify a different model on the fly which is not the case. 

I think we need to improve documentation more before we merge this.

jaresty:
You can't override this locally unless you modify the repository-it also requires modifying in several places. I really wanted to experiment with some different model names-I find the word model really difficult to say. Can we improve the documentation to make this possible to merge?

jaresty:
Fwiw, I think there is a possible future feature to support multiple models in the fly as well since there are various reasons to prefer one or the other and it's currently difficult to switch here to experiment. The continue extension, for example, does support switching in the fly. I don't want to expand the scope of this PR but I wanted to point out that we may need some grammar to switch eventually.

jaresty:
To give you a sense of some of the names I'm considering:
- skip
- brain
- jeep
- pete
- bird
- dog
- ...


I think it's obvious why some of these aren't sensible to upstream. I'm using "model" as often or more often than I'm using cursorless commands and even letter names, so having some way to customize the name is really becoming an issue for me. Can you think of another way to customize the name that does not require forking?

C-Loftus:
I am ok with merging this and compromising if we can get rid of the model modifiers list as you mentioned in slack. Not sure if you have seen the docs site and how we are generating it. My goal is to have a docs as code setup. At build time the docs directly fetch the source files and output them in the docs. Hence why I want to have very clear and annotated `.talon` files. This makes it so our docs are hopefully always in sync with the state of the repo. The more modifiers / lists we have in the commands, the harder it is for a new user to understand how to use the command. Does that seem reasonable where I am coming from? I still am trying to focus on improving UX for new users.

jaresty:
Totally! I'm going to get started on replacing the modifiers with destinations next. I think that will be a good improvement.

jaresty:
I think I wrapped up removing the modifiers - let me know if you see any issues. I updated this pull request based on the new changes, so hopefully this is now ready to review.

C-Loftus:
Looks good to me with two minor comments.

1. I wonder if should just use `user.model` as the name for the list. Or perhaps `user.modelName` or `user.modelKeyword`. Honestly not sure if these are more clear. Trying to think which way of structuring it is most clear for a new user who will be reading the `.talon` file and needs to understand that the keyword `model` is a part of that list. 
2. I added this documentation to the docs site and removed it from the readme. I think we can probably just use the readme for very essential info and otherwise put everything else in the docs website.

jaresty:
I like the idea of "user.model" as the name- then the default reads the same as before.

C-Loftus:
Looks good to me. And in the future if needed we can switch to `modelName` if users fine this confusing or it conflicts with another list for some reason.

PR #137, opened 2024-08-06
Fixed bug in model blend paste

jaresty:
- it also needs to extract the response


jaresty:
I'm going to merge this one since it fixes a regression.

PR #140, opened 2024-08-07
Removing thread from modifiers

jaresty:
- We're trying to simplify the grammar
- as the first step in removing the thread from a modifier is to make it available as a setting that can be enabled


jaresty:
there's still some work to do on removing it - I'll keep working on that later

jaresty:
I thought there was more work to do, but it seems all I had to do was remove it from the modifier list.

jaresty:
Once this is merged I can remove snip and remove the modifier list as well 🔥

C-Loftus:
Good with me. Feel free to merge any of these minor changes we've discussed. As long as it doesn't regress behavior, change user-facing things in a significant way we haven't discussed, or cause a significant internal code refactor, you are good to merge 👍

PR #141, opened 2024-08-08
Remove snip modifier and model modifiers list

jaresty:
- Snip is now a destination

### DIFF
--- a/GPT/lists/modelDestination.talon-list
+++ b/GPT/lists/modelDestination.talon-list
@@ -41,3 +41,6 @@ to window: window
 
 # Select the response after insertion so you can apply subsequent prompts on it
 chain: chain
+
+# Insert the response as a snippet with placeholders (only works in vscode)
+snip: snip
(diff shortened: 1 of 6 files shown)


GPT/lists/modelDestination.talon-list
```diff
@@ -41,3 +41,6 @@ to window: window
 
 # Select the response after insertion so you can apply subsequent prompts on it
 chain: chain
+
+# Insert the response as a snippet with placeholders (only works in vscode)
```
C-Loftus, line 45: Useful callout  :+1:

lib/modelHelpers.py
```diff
@@ -187,12 +187,3 @@ def get_clipboard_image():
     except Exception as e:
         print(e)
         raise Exception("Invalid image in clipboard")
-
-
-def paste_and_modify(formatted_message: str, modifier: str = ""):
```
C-Loftus, line 192: Nice cleanup!

PR #142, opened 2024-08-08
First pass at tracking thread roles

jaresty:
- we want to distinguish between user and assistant

### DIFF
--- a/lib/modelHelpers.py
+++ b/lib/modelHelpers.py
@@ -129,7 +129,7 @@ def send_request(
 
     language = actions.code.language()
     language_context = (
-        f"\nThe user is currently in a code editor for {language}."
+        f"The user is currently in a code editor for the programming language: {language}."
         if language != ""
         else None
     )
--- a/lib/modelHelpers.py
+++ b/lib/modelHelpers.py
@@ -1,7 +1,7 @@
 import base64
 import json
 import os
-from typing import Optional, Tuple
+from typing import Optional
 
 import requests
 from talon import actions, app, clip, settings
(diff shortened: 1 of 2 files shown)


jaresty:
I'm going to use this for a while, but I think it's working now.

jaresty:
I think this is working pretty well now. Let me know if you have any feedback or if you'd like me to merge it-I think it would be good to get this in before any other big changes.

jaresty:
Importantly this introduces a difference in the way that we are storing threads versus context:

1. context continues to be stored as an array of messages 
2. threads are now stored as an array of messages at one layer up including the roles

This is because context is now always part of the system role, while threads vary from assistant and user. A future refactor might look at making the context stored in the same way however we build up system context with every request so it might be fine just as is. Tracking whether the messages are coming from the assistant, the user, or in the system context more accurately seems to have a subtle yet good impact on the responses I'm getting.

lib/modelHelpers.py
```diff
@@ -93,12 +109,12 @@ def format_clipboard():
         return format_message(clip.text())
 
 
-def generate_payload(
+def send_request(
```
C-Loftus, line 112: We still have a `gpt_send_request` function so I think the similar names of both `send_request` and `gpt_send_request` is a bit confusing. If we are just calling `gpt_send_request` from `send_request` and the former is a small function, we can probably just move it inside.

C-Loftus:
Generally looks good to me with a comment about the naming of fns. May make some clean up edits in a new PR later today after this is merged.

lib/modelHelpers.py
C-Loftus: That being said I might be missing something since I didn't look through the entirety of the codebase in relation to this change.

jaresty: moved

C-Loftus (approved):

PR #144, opened 2024-08-09
Refactored to use typed dicts. Removed stale code

C-Loftus:
Kept same behavior the same as far as I can tell.

- Removed all dict with `any` as a key or value. Try to use `GPTMessage` or `GPTMessageItem` as the type as much as possible. Let me know if there is a name you prefer instead of these. I wasn't sure what to name them.
- Got rid of old tts and image description code that I don't think is relevant anymore. These don't follow the grammar and aren't being used since they were broken but no one reported an issue
-  No pylance errors are being thrown under `standard` type checking
- Got rid of the a11y editor context logic in the system prompt. I will probably add something to make that more explicit. I don't like that it would be always passed in if the user has screen reader support enabled for axkit
- Use the word `{user.model} **start** thread:` since it is one syllable instead of `enable` which is 3 syllables.
- Changed `contextual_user_context` to be `gpt_additional_user_context` since it is a globally exposed talon python function, we want to prefix it with `gpt`. I personally find `additional_user_context` to be more clear than `contextual_user_context` which strikes me as a bit confusing.

I am using `TypeDict` so we keep all the methods within the standard `dict` class. Was considering creating a super class for the message and inheriting from it but I think this abstraction does not actually benefit and adds boilerplate.


C-Loftus:
Have not tested this yet. Feel free to test if you want. Going to take a break for a bit.

C-Loftus:
Believe this should be good now. The threading is very cool thank you for your work on this. I find it works well. I will add funcitonality to have the imgui window auto-update with a new thread response once this merged.

GPT/gpt.talon
```diff
@@ -5,7 +5,7 @@
 #   Example: `model fix grammar below` -> Fixes the grammar of the selected text and pastes below
 #   Example: `model explain this` -> Explains the selected text and pastes in place
 #   Example: `model fix grammar clip to browser` -> Fixes the grammar of the text on the clipboard and opens in browser`
-{user.model} <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]:
+{user.model} <user.modelPrompt> [{user.modelSource}] [{user.modelDestination}]$:
```
C-Loftus, line 8: Added this. Let me know if you disagree but I think there are few situations in which a user will want to chain the model command since they haven't seen the output until it is returned. I think $ prevents a lot of the weird situations where talon thinks there is a second command inside of this.

This still allows for prefix chaining before the command

jaresty: I was thinking about going further and including the ^ before the start of the command. I didn't understand what prefix chaining is-can you give me an example?

jaresty:
I'll use this branch for a while. Looks good though!

GPT/gpt.talon
C-Loftus: When I say prefix chaining I just mean chaining another talon command before. So for instance you could use a cursory list command before and then chain with the model as long as it has at the end of the chain

ie 'take block this model fix grammar'

All as one sentence. That is something I sometimes use.

jaresty: Ah I see. In that particular case I suppose you could use the cursorless beta command too (I.e. model fix grammar block) but yeah I can see what you mean.

C-Loftus: Yup, although for arbitrary commands that don't fit in the grammar then it becomes a little slower to use. I don't think anything is disrupted by not having the anchor at the start so I think we can just have the anchor at the end to prevent chaining after. The main issue is having the ability to chain after user.text. (Sometimes Talon does not know if the user is trying to dictate raw text or break out of that with a separate command)

The $ end anchor removes that so it is probably sufficient just like this

jaresty (approved):

PR #147, opened 2024-08-10
Better model window for threading

C-Loftus:
- renamed confirmation gui functions so they have proper naming style
- Made the text that we operate on optional as a param. We were previously passing empty user messages with "" as text. If it is none, we want to ignore this and not add it.
- Added a window for the thread. If you have the thread started and the window is open, it will auto update the window with the new thread data.
- If you don't specify a model destination when the window is open, it will always add to the window
- added a model toggle window function to see the thread (I am ok with getting rid of this, not sure if it is proper if we already have model pass thread to window. The issue with the latter is that it is difficult to know when it is coming from the thread in order to format it properly


lib/modelHelpers.py
```diff
@@ -142,9 +140,12 @@ def send_request(
 
     current_request: GPTMessage = {
         "role": "user",
-        "content": [prompt, content],
+        "content": [prompt],
     }
 
+    if text_to_process is not None:
+        current_request["content"].append(text_to_process)
+
```
C-Loftus, line 148: in a future PR I think we should probably consolidate the text that is operated on and the prompt all into one message. I don't think the model needs both as separate messages and it doesn't make as much sense semantically to split them if we are referring to content that is in a different message. ie `please explain this` `foo bar` seems worse than `please explain this: \n\n foo bar` since `this` could be ambiguous with regards to which message. I agree though that system messages should still be split into separate messages.

jaresty:
Excited to try this out-won't be able to during the day though.

jaresty:
I tried this out and I like it. I wonder if the windows should include the user prompts as well, or if not-maybe some kind of delimiter to show when the next thread starts.

C-Loftus:
> I tried this out and I like it. I wonder if the windows should include the user prompts as well, or if not-maybe some kind of delimiter to show when the next thread starts.

Did you pull the latest version? The integration last I pushed does show both the user prompt and the assistant response with a newline between the two

jaresty:
I see it now. I think I was confused because the first time it didn't show (when I just said 'model ask what can I do to window'). The subsequent messages were showing the roles. They're pretty subtle though-easy to lose track of in all the text of the responses. I wonder if there's something that we could do to make them more prominent. At any rate it's a really nice improvement-am fine with merging this now.

jaresty (approved):

C-Loftus:
So the simplest thing would be to simply make the role all caps. Fine with that if you'd want that. 

But I agree that I would like to make things more clear for users in general even beyond the roles. Imgui is honestly a pretty limited tool. I think having fewer buttons could reduce screen space or adding more \n\n could help. I still sorta feel like we probably don't need model chain and model pass to context on there. Or at least not when it is in the thread view. Perhaps we can take those out in that case specific to the thread.

But yeah not sure how else to make general improvements since imgui doesn't support horizontal layouts and is pretty limited, but it is all Talon supports for self-contained GUIs.

jaresty:
I think I'm okay with removing model chain; I guess we could remove pass to context for now as well. I might want to revisit that in the future but we can remove it for now. I think this is good.

C-Loftus:
Merging this. Took out the extra buttons for chain and pass to context since I think for most users this is overkill and clutters the screen. However, the voice command is still there so they can still say it or use the standard pass grammar.

Cleaned up the window so we have all caps for USER / GPT and the window displays differently depending on whether it is displaying a thread or raw text.

C-Loftus:
Also split the context commands into a new file so it is easier for docs as code

PR #149, opened 2024-08-11
Provide a setting for default destination

jaresty:
- Sometimes you don't want the default to be paste
- In zoom calls for example, pasting does not make much sense
- By providing a setting that can be overridden with a talon setting users can set a different default per application if they want


C-Loftus (approved):
Fine with me. Some potential for feature creep with additional settings but probably ok

PR #152, opened 2024-08-11
Wrapped content to distinguish prompt and content

jaresty:
- Even those these end up as separate chunks, by clearly wrapping the text in delimiters we are able to distinguish the prompt from the content better


jaresty:
Closing in favor of https://github.com/C-Loftus/talon-ai-tools/pull/153

PR #153, opened 2024-08-11
Here is an alternative refactor for content wrapping

jaresty:
- this lets us avoid changing the format messages calls everywhere


C-Loftus:
Looks good to me. Just as an fyi we need to handle the case where the content is empty for instance when the user says a model please command without specifying a target and just wants to generate new text. 

I am going to hold off on merging this until I merge the other pr since it contains a bit of logic to make this easier to reason about

C-Loftus:
Closing since addressed in other PR a bit more robustly

PR #155, opened 2024-08-12
When the thread is empty and you try to open a window this throws an error

jaresty:


C-Loftus (approved):
Thank you for catching this 👍

PR #156, opened 2024-08-12
Image handling is currently broken

jaresty:
- When you fetch the image source it formats the messages
- If you try to extract them as text at that point, images will not be processed at all
- In order to process images you have to keep the messages intact from source extraction through to insertion

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -137,17 +137,17 @@ class UserActions:
         """Apply an arbitrary prompt to arbitrary text"""
 
         text_to_process: GPTMessageItem = actions.user.gpt_get_source_text(source)
+        if (
+            text_to_process.get("text", "") == ""
+            and text_to_process.get("image_url", "") == ""
+        ):
+            text_to_process = None  # type: ignore
 
         # Handle special cases in the prompt
         ### Ask is a special case, where the text to process is the prompted question, not selected text
         if prompt.startswith("ask"):
             text_to_process = format_message(prompt.removeprefix("ask"))
             prompt = "Generate text that satisfies the question or request given in the input."
-        elif (
-            text_to_process.get("text", "") == ""
-            and text_to_process.get("image_url", "") == ""
-        ):
-            text_to_process = None  # type: ignore
 
         response = gpt_query(format_message(prompt), text_to_process, destination)
 
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -207,6 +207,21 @@ class UserActions:
             # Skip inserting the response if the user is just viewing the thread in the window
             actions.user.confirmation_gui_refresh_thread()
 
+        match method:
+            case "thread":
+                GPTState.push_thread(format_messages("user", [gpt_message]))
+                return
+            case "newThread":
+                GPTState.new_thread()
+                GPTState.push_thread(format_messages("user", [gpt_message]))
+                return
+
+        if gpt_message.get("type") != "text":
+            actions.app.notify(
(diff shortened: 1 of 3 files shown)


GPT/gpt.py
```diff
@@ -138,27 +138,22 @@ def gpt_apply_prompt(prompt: str, source: str = "", destination: str = ""):
 
         text_to_process: GPTMessageItem = actions.user.gpt_get_source_text(source)
 
-        # If after creating the message there wasn't anything in the text_to_process, set it to None so
```
jaresty, line 141: This check was blocking images from being processed

GPT/gpt.py
```diff
@@ -212,72 +207,75 @@ def gpt_insert_response(
             # Skip inserting the response if the user is just viewing the thread in the window
             actions.user.confirmation_gui_refresh_thread()
 
+        message_text_no_images = extract_message(gpt_message)
```
jaresty, line 210: Extract message currently returns text so if there's an image here it will return an empty string

lib/modelState.py
```diff
@@ -40,6 +40,11 @@ def disable_thread(cls):
     @classmethod
     def push_context(cls, context: GPTMessageItem):
         """Add the selected text to the stored context"""
+        if context.get("type") != "text":
```
jaresty, line 43: This ensures that images don't get inserted into the system messages since that would cause an error

GPT/gpt.py
```diff
@@ -138,27 +138,22 @@ def gpt_apply_prompt(prompt: str, source: str = "", destination: str = ""):
 
         text_to_process: GPTMessageItem = actions.user.gpt_get_source_text(source)
 
-        # If after creating the message there wasn't anything in the text_to_process, set it to None so
-        # we don't send meaningless data to the model
-        if text_to_process.get("text", "") == "":
-            text_to_process = None  # type: ignore
```
C-Loftus, line 144: Unless I am missing other change, we still need to set this here since otherwise it will pass empty messages when using `model please`. Updated the check to check for both images and text and only then setting it to be None.

```suggestion
       if text_to_process.get("text", "") == "" and text_to_process.get("image_url") == "":
            text_to_process = None  # type: ignore
```

jaresty: There is one issue with that - when using a thread you won't always have text you are trying to modify.  You could be asking followup questions about the context that is already in the thread, in which case there would not be any text to modify.  I regularly add things to the context or ask questions with no target text.  Sometimes I do it just to ask a question about general knowledge (i.e. who is Kevin Bacon?)

jaresty: Maybe model please should be a special case that requires the text to be present.

C-Loftus: Sorry I am not following your comment. `text_to_process` is just a variable that is passed through and eventually gets concatenated to the prompt if it is not `None`. If the user didn't pass in any text or an image we don't want to concatenate an empty message and we can ignore this variable. I don't think this affects threading at all. 

We still can use the thread or context if this variable is None.

jaresty: I couldn't apply your change for some reason, but I did it in the next commit for all prompts other than 'ask'.

jaresty: oh I see what you mean-my bad. I'll change it back

C-Loftus (approved):
Made two quick edits. If the user wanted to append to the thread, we want to refresh the GUI. The gui refresh function won't do anything if it is closed so its fine either way.

thanks for help cleaning this up

PR #158, opened 2024-08-14
Add debug logging functionality for GPT interactions

jaresty:
- Introduced methods to enable and disable debug logging in UserActions class.
- Added Talon commands to toggle debug logging.
- Updated modelHelpers.py to print request data when debug is enabled.
- Modified GPTState class to include debug_enabled flag and methods to toggle debug state.

### DIFF
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -91,6 +91,14 @@ class UserActions:
             "text", ""
         )
 
+    def gpt_enable_debug():
+        """Enable debug logging"""
+        GPTState.enable_debug()
+
+    def gpt_disable_debug():
+        """Disable debug logging"""
+        GPTState.disable_debug()
+
     def gpt_clear_context():
         """Reset the stored context"""
         GPTState.clear_context()
--- a/GPT/gpt.py
+++ b/GPT/gpt.py
@@ -91,13 +91,13 @@ class UserActions:
             "text", ""
         )
 
-    def gpt_enable_debug():
+    def gpt_start_debug():
         """Enable debug logging"""
-        GPTState.enable_debug()
+        GPTState.start_debug()
 
-    def gpt_disable_debug():
+    def gpt_stop_debug():
         """Disable debug logging"""
-        GPTState.disable_debug()
+        GPTState.stop_debug()
 
     def gpt_clear_context():
         """Reset the stored context"""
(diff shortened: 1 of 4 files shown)


C-Loftus:
I am fine with this, but if there are only two states, enabled or disabled, and the methods don't have other side effects, I think we should just have one toggle method. And then the command would be model toggle debug. Could you make that change?

GPT/gpt.talon
```diff
@@ -23,3 +23,9 @@
 {user.model} [nope] that was <user.text>$:
     result = user.gpt_reformat_last(text)
     user.paste(result)
+
+# Enable debug logging so you can more details about messages being sent
+{user.model} enable debug: user.gpt_enable_debug()
+
+# Disable debug logging
+{user.model} disable debug: user.gpt_disable_debug()
```
C-Loftus, line 31:
```suggestion
# Toggle debug logging so you can more details about messages being sent
{user.model} toggle debug: user.gpt_enable_debug()
```

GPT/gpt.py
```diff
@@ -91,6 +91,14 @@ def gpt_generate_sql(text_to_process: str) -> str:
             "text", ""
         )
 
+    def gpt_enable_debug():
+        """Enable debug logging"""
+        GPTState.enable_debug()
+
+    def gpt_disable_debug():
+        """Disable debug logging"""
+        GPTState.disable_debug()
```
C-Loftus, line 100:
```suggestion
    def gpt_toggle_debug():
        """Toggle debug logging"""
        GPTState.toggle_debug()
```

jaresty:
To be honest I prefer to have explicit commands to enable and disable in this case because it's not easy to view the current debug state so it would be easy to get confused about which state it is in. I would be okay with changing the commands to be 'start' and 'stop' though if that's better.

C-Loftus:
OK let's use start and stop then just to keep things consistent. Then you can merge after that

lib/modelState.py
```diff
@@ -12,6 +12,19 @@ class GPTState:
...
+    def enable_debug(cls):
+        """Enable debug printing"""
+        cls.context = []
+        actions.app.notify("Enabled debug logging")
+
+    @classmethod
+    def disable_debug(cls):
+        """Disable debug printing"""
+        cls.context = []
+        actions.app.notify("Disabled debug logging")
```
C-Loftus, line 27: Now that I am looking at this again, where is `debug_enabled` being toggled? Don't we also want to keep context the same? I think implicitly clearing it is confusing if the user wants to debug

jaresty: Oops that is a mistake-will fix

C-Loftus:
Looks good now, thanks!

PR #162, opened 2024-09-25
Fix Header Handling for Azure-Specific Authorization

### LINKED ISSUE #161: Azure endpoints require a different header
Hi there, 

This project is awesome, thanks for all your work on this. 

Just FYI, it works with Azure OpenAI endpoints, just with a slight tweak to the header. 

https://github.com/C-Loftus/talon-ai-tools/blob/e98c2d0577ddfbe3f210bcb60de6fcc38bd2e8c0/lib/modelHelpers.py#L138
needs to become 
```python
"api-key": TOKEN,
```

Not sure if that's something you want to do dynamically or if just leaving this issue for other people to find and tweak themselves is enough :) I imagine supporting every single model end point would be a lot of work

Max-FM:
# Description
This pull request fixes #161 by updating header management for Azure-specific authorization requirements in both the `lib/modelHelpers.py` and `Images/ai-images.py` files.

## Changes

1. **lib/modelHelpers.py:**
   - **Moved header setup:** The Authorization and Content-Type headers were moved closer to the request execution.
   - **Conditional logic for Azure endpoints:** For Azure endpoints, the api-key header is used instead of Authorization. Non-Azure endpoints continue using the Authorization header.

2. **Images/ai-images.py:**
   - **Modified header setup:** Added conditional logic to handle the api-key header for Azure and Authorization for OpenAI.

## Note
While the headers have been updated for Azure compatibility, image generation will still not function with Azure, as the endpoint remains hardcoded to OpenAI. A future PR will be required to support Azure image generation.

### DIFF
--- a/Images/ai-images.py
+++ b/Images/ai-images.py
@@ -15,10 +15,12 @@ class Actions:
 
         url = "https://api.openai.com/v1/images/generations"
         TOKEN = get_token()
-        headers = {
-            "Content-Type": "application/json",
-            "Authorization": f"Bearer {TOKEN}",
-        }
+        headers = {"Content-Type": "application/json"}
+        # If the model endpoint is Azure, we need to use a different header
+        if "azure.com" in url:
+            headers["api-key"] = TOKEN
+        else:
+            headers["Authorization"] = f"Bearer {TOKEN}"
         data = {
             "model": "dall-e-3",
             "prompt": prompt,
--- a/Images/ai-images.py
+++ b/Images/ai-images.py
@@ -19,6 +19,7 @@ class Actions:
         # If the model endpoint is Azure, we need to use a different header
         if "azure.com" in url:
             headers["api-key"] = TOKEN
+       # otherwise default to the standard header format for openai
         else:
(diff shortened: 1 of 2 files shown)


Images/ai-images.py
```diff
@@ -15,10 +15,12 @@ def image_generate(prompt: str):
...
-        headers = {
-            "Content-Type": "application/json",
-            "Authorization": f"Bearer {TOKEN}",
-        }
+        headers = {"Content-Type": "application/json"}
+        # If the model endpoint is Azure, we need to use a different header
+        if "azure.com" in url:
+            headers["api-key"] = TOKEN
+        else:
+            headers["Authorization"] = f"Bearer {TOKEN}"
```
C-Loftus, line 23:
```suggestion
       # otherwise default to the standard header format for openai
        else:
            headers["Authorization"] = f"Bearer {TOKEN}"
```

C-Loftus:
Thanks! Don't worry too much about image generation (fine either way though!) unless you are using it and want it. I think it's more of a novelty; wasn't even sure if it worked with Azure hence the hardcoding.

PR #166, opened 2025-02-06
Fix copilot_inline_chat to always start inline chat.

### LINKED ISSUE #12: "pilot change" breaks if there are other rewrite actions available
See https://github.com/orgs/community/discussions/103149. I believe this is out of our control for now, but once that discussion is addressed we should be able to just switch to the more specific action kind here

wolfmanstout:
Fixes #12


C-Loftus:
Thank you!

PR #167, opened 2025-02-23
Wrap long lines in the window.

wolfmanstout:
Configurable with user.model_window_char_width.

### DIFF
--- a/lib/modelConfirmationGUI.py
+++ b/lib/modelConfirmationGUI.py
@@ -1,4 +1,6 @@
-from talon import Context, Module, actions, clip, imgui
+import textwrap
+
+from talon import Context, Module, actions, clip, imgui, settings
 
 from .modelHelpers import GPTState, extract_message, notify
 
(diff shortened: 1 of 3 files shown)


C-Loftus (commented):
Looks good to me, thanks! Could we add a sample line in [the example settings file](https://github.com/C-Loftus/talon-ai-tools/blob/main/talon-ai-settings.talon.example) just so this is documented? (Nice thing is that the docs site also auto fetches this content at build time so it will also be documented on the website if we add this)

wolfmanstout:
> Looks good to me, thanks! Could we add a sample line in [the example settings file](https://github.com/C-Loftus/talon-ai-tools/blob/main/talon-ai-settings.talon.example) just so this is documented? (Nice thing is that the docs site also auto fetches this content at build time so it will also be documented on the website if we add this)

Done.

C-Loftus (approved):

jaresty:
Thanks! Cherry picked to my fork.

PR #168, opened 2025-02-28
Implemented simonw/llm backend.

wolfmanstout:


lib/modelHelpers.py
```diff
@@ -119,21 +119,22 @@ def send_request(
...
     )
 
-    system_messages: list[GPTMessageItem] = [
-        {"type": "text", "text": item}
-        for item in [
-            settings.get("user.model_system_prompt"),
-            language_context,
-            application_context,
-            snippet_context,
+    system_message = "\n\n".join(
```
wolfmanstout, line 122: Note that this is the only small change that will also affect current behavior: joining the system prompts into one string. I thought it better to standardize on something that could be supported by both backends. Multiple system messages is not a common pattern regardless.

lib/modelHelpers.py
```diff
@@ -189,21 +213,64 @@ def send_request(
...
+    cmd_input: bytes | None = None
+    if content_to_process and content_to_process["type"] == "image_url":
+        img_url: str = content_to_process["image_url"]["url"]  # type: ignore
+        if img_url.startswith("data:"):
+            command.extend(["-a", "-"])
+            base64_data: str = img_url.split(",", 1)[1]
+            cmd_input = base64.b64decode(base64_data)
+        else:
+            command.extend(["-a", img_url])
+    model: str = settings.get("user.openai_model")  # type: ignore
```
wolfmanstout, line 240: We should probably deprecate this setting and introduce a new one, e.g. model_default_model (considering that I have future changes planned that allow the user to use a different model based on the utterance). Alternatively we could deprecate and not add a new setting and simply have "model" map directly to the model the user chooses, which they would override with a custom .talon-list.

lib/talonSettings.py
```diff
@@ -53,7 +53,14 @@ def modelSimplePrompt(matched_prompt) -> str:
     "model_endpoint",
     type=str,
     default="https://api.openai.com/v1/chat/completions",
-    desc="The endpoint to send the model requests to",
+    desc='The endpoint to send the model requests to. If "llm" is used, the llm CLI is used (see https://github.com/simonw/llm).',
```
C-Loftus, line 56:
```suggestion
    desc='The endpoint to send the model requests to. If "llm" is specified instead of a url, the llm CLI tool is used when routing all language model requests (see https://github.com/simonw/llm).',
```

talon-ai-settings.talon.example
```diff
@@ -7,6 +7,11 @@ settings():
     # Works with any API with the same schema as OpenAI's (i.e. Azure, llamafiles, etc.)
     # user.model_endpoint = "https://api.openai.com/v1/chat/completions"
```
C-Loftus, line 8:
```suggestion
    # Set the endpoint that the model requests should go to. 
    # Works with any API with the same schema as OpenAI's (i.e. Azure, llamafiles, etc.)
    # Set to "llm" to use the local llm cli tool as a helper for routing all your model requests
    # user.model_endpoint = "https://api.openai.com/v1/chat/completions"
```

readme.md
```diff
@@ -12,17 +12,39 @@ This functionality is especially helpful for users who:
 **Prompts and extends the following tools:**
 
 - Github Copilot
-- OpenAI API (with any GPT model) for text generation and processing
+- OpenAI API (with any GPT model) or [simonw/llm CLI](https://github.com/simonw/llm) for text generation and processing
   - Any OpenAI compatible model endpoint can be used (Azure, local llamafiles, etc)
 - OpenAI API for image generation and vision
```
C-Loftus, line 17: While we are editing the readme might as well remove this old info
```suggestion
- OpenAI API for image recognition
```

lib/modelHelpers.py
C-Loftus: Fine with me. Agreed that standardizing is better and I don't think this makes a significant difference

C-Loftus (commented):
Looks good to me; added some comments, mainly documentation stuff that should be easy to resolve. I tested this a bit, not super thoroughly but seems fine to me.

Would be super nice if there was a way to test this but might be too annoying to mock talon libs.

lib/modelHelpers.py
C-Loftus: Agreed

wolfmanstout: Any preference on new setting vs just use the talon list? Is a follow up PR okay or you want to fix here?

C-Loftus: Not entirely sure to be honest. I like a setting since it is simpler and easier to override but I like the talon list since it is probably less config for the user to maintain or switch model choices. 

Happy to defer to your preference. If we use the talon list we just should make it super clear what the default model is, analogous to what the default model would be for the setting value, and also make it super clear how the talon list interacts with llm

readme.md
```diff
@@ -34,12 +37,42 @@ import os
...
+# user.model_llm_path = "/path/to/llm"
+```
+
+3. Choose a model in settings:
+
+```
+user.openai_model = "claude-3.7-sonnet"  # or whichever model you installed
+```
+
+4. By default, all model interactions will be logged and viewable via `llm logs`. If you prefer, you can disable this with `llm logs off`.
```
C-Loftus, line 57:
```suggestion
4. By default, all model interactions will be logged locally and viewable on your machine via `llm logs`. If you prefer, you can disable this with `llm logs off`.
```

readme.md
```diff
@@ -34,12 +37,42 @@ import os
 os.environ["OPENAI_API_KEY"] = "YOUR-KEY-HERE"
 ```
 
-5. See the [GPT](./GPT/readme.md) or [Copilot](./copilot/README.md) folders for usage examples.
+### Option 2 (recommended): simonw/llm CLI
+
+1. Install [simonw/llm](https://github.com/simonw/llm#installation) and set up one or more models to use.
```
C-Loftus, line 42:
```suggestion
1. Install [simonw/llm](https://github.com/simonw/llm#installation) and set up one or more models to use.

Run `llm keys set` with the name of the provider you wish to use to set the API key for your requests. i.e. `llm keys set openai`.
```

C-Loftus (commented):
Looks good. Two other quick edits on docs if you think it is reasonable. Figure we should probably list how to set the API key for llm

wolfmanstout:
I think this is good to go. I improved the formatting of one of your suggested changes (I think the bit about the API key is more of a note than a required step considering that llm can work with local models that don't require keys).

C-Loftus:
Sounds good merging. Mainly wanted to call out API key since I think most people will use managed services with llm but I could be wrong.

PR #169, opened 2025-03-05
Add model selection within utterances.

wolfmanstout:
Enhances talon commands to allow specifying a model directly within the utterance. For example: 'four o mini explain this' will use gpt-4o-mini model.

- Updated model.talon-list to include four o and four o mini models. I only included OpenAI models by default because it would be a significant maintenance effort to support every model supported by simonw/llm.
- Modified send_request and supporting functions to accept and use model parameter
- Updated all Talon command files to pass the model parameter
- Fixed parameter ordering to ensure optional params come after required ones
- Deprecated openai_model setting in favor of model_default_model

Note: I tested every command here. I fixed some issues with the "model find" command. I noticed that the Cursorless commands are broken, and there are some bugs in the beta "blend" commands. I compared to the baseline and these issues were already present. I can file separate bugs for those, since they seem to run a bit deeper.


GPT/lists/model.talon-list
```diff
@@ -3,3 +3,6 @@ list: user.model
 
 # The default model name
 model: model
+# OpenAI models
```
wolfmanstout, line 6: Where do you want me to document that the user can add to this list if they want to add more models (e.g. supported by simonw/llm)? I could put this in either the top-level readme or the gpt readme. They would need to modify this list directly or else I suppose they could create their own list with a more specific context (e.g. following something like this: https://github.com/talonhub/community?tab=readme-ov-file#-tip-overriding-cleanly). IMO the latter recommendation is kind of awkward and verbose vs. simply changing a file, but I could put something similar if you want.

C-Loftus:
Thanks for this PR, may be a couple days until I properly review/test locally just fyi; busy next few days; feel free to ping if I forget

GPT/lists/model.talon-list
C-Loftus: If it works for you, it would be great to have a sentence of two copy-pasted directly in this .talon-list file itself. But yeah I don't think we need to worry about this too much since it is more a talon specific question regardless.

lib/talonSettings.py
```diff
@@ -30,9 +28,17 @@ def modelSimplePrompt(matched_prompt) -> str:
 
 
 mod.setting(
-    "openai_model",
-    type=Literal["gpt-3.5-turbo", "gpt-4", "gpt-4o-mini"],  # type: ignore
+    "model_default_model",
```
C-Loftus, line 31: I am neutral on the setting `model_default_model`. I think it is sort of confusing to some users who don't come from a coding background and might not pick up that `model_` is the namespace. I am wondering if we just want to use `user.model_default` or if that is too brief. In retrospect I probably should've just namespaced the entire repo's settings with `user.ai_tools_X` but thats ok.

Happy to defer to what you think is best.

GPT/beta-commands/beta-gpt.py
```diff
@@ -11,7 +12,7 @@
 
 @mod.action_class
 class UserActions:
-    def gpt_find_talon_commands(command_description: str):
+    def gpt_find_talon_commands(command_description: str, model: str):
```
C-Loftus, line 15: Thanks for checking these functions. Thinking out loud, I am wondering if maybe I should just remove the gpt beta stuff. I don't think it is commonly used and stuff like the find_talon_commands functionality is probably best served for AI driven docs like the ones you made.

Will keep for now but worth revisiting.

GPT/gpt.py
```diff
@@ -55,15 +57,19 @@ def gpt_blend(source_text: str, destination_text: str) -> None:
...
+    ) -> None:
         """Blend all the source text as a list and send it to the destination"""
 
-        return actions.user.gpt_blend("\n---\n".join(source_text), destination_text)
+        return actions.user.gpt_blend(
+            "\n---\n".join(source_text), destination_text, model
+        )
 
-    def gpt_generate_shell(text_to_process: str) -> str:
+    def gpt_generate_shell(text_to_process: str, model: str) -> str:
```
C-Loftus, line 72: Totally good w me either way but curious on the rationale for passing in model to each function instead of just calling the setting for it since it is global state anyways. I suppose this is better if we ever set up test w/ mocking?

C-Loftus (commented):
Thank you for the PR and thank you for your patience while I was away w/ travel. Made a few general comments, but it all looks good. Once aligned we can merge.

GPT/lists/model.talon-list
wolfmanstout: Done.

GPT/beta-commands/beta-gpt.py
wolfmanstout: Sure, that's up to you.

GPT/gpt.py
wolfmanstout: I have to pass in the model in order to support models provided within the utterance. With this change, anywhere that you can say "model" you can now also name a specific model to use. The global state will always be the same, but this parameter passes along the list value from the utterance.

wolfmanstout:
I've addressed all of your comments.

C-Loftus:
Looks good, thanks!

PR #172, opened 2025-03-26
Fix model resolution to ensure consistent model name handling

wolfmanstout:
Added a centralized resolve_model_name function in modelHelpers.py to ensure that all commands using {user.model} properly resolve to the default model when the placeholder "model" is used.

This wasn't working for "model shell", for example.


C-Loftus (approved):
Looks reasonable to me!

PR #175, opened 2025-03-30
Remove threading code and clean up gpt beta

C-Loftus:
- Remove all threading code; this is best handled with the llm cli; managing persistent global state inside Talon is messy and error prone to maintain
- Get rid of the gpt-beta tag and either move out the useful functions or get rid of the less useful ones like 
    - experiments like trying to pass in all talon files into an llm for dynamic help is a good idea, but in reality didn't work very well
    - Some of this might be worth experimenting with again at some point, but for the time being, probably best to remove in an effort to clean things up 
 - Get rid of gpt blend functions from the main codebase
     - Stuff that calls core gpt functions in an interesting ways are very cool, but at least at this point, probably belongs moreso in a personal config for the purposes of maintainability. Talked with James and sometimes it can be easy to miss cases like these functions that may be less frequently used.


C-Loftus:
No rush on review at all; just wanted to clean things up in an effort to make the codebase a little cleaner for you to PR against. 

Haven't tested this super thoroughly but aiming for this PR to not change anything in the core behavior; just that threading logic / removal of legacy stuff.

wolfmanstout (approved):

PR #176, opened 2025-03-30
Add thread continuation support through 'and' keyword (continueLast)

wolfmanstout:
- Add thread.talon-list with 'and: continueLast' to continue the most recent conversation
- Add thread parameter to all model functions and commands
- Display warning when thread continuation is attempted without the llm CLI
- Support the thread parameter everywhere {user.model} is used

Thread continuation works with all model commands, e.g.:
- 'model and explain this'
- 'four o mini and shell list all directories'

This feature currently requires setting user.model_endpoint = 'llm'

### DIFF
--- a/GPT/gpt-cursorless.talon
+++ b/GPT/gpt-cursorless.talon
@@ -4,9 +4,9 @@ tag: user.cursorless
 
 # Apply a prompt to any text, and output it any target
 # Paste it to the cursor if no target is specified
-{user.model} <user.modelSimplePrompt> <user.cursorless_target> [<user.cursorless_destination>]$:
+{user.model} [{user.thread}] <user.modelSimplePrompt> <user.cursorless_target> [<user.cursorless_destination>]$:
     text = user.cursorless_get_text_list(cursorless_target)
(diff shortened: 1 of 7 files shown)


C-Loftus:
Looks good to me. Just wanted to clarify the reasoning behind having the thread command in a talon list with one item. Is that so it can be overriden and customized? Or do we intend on one day allowing for more complex thread specifiers beyond just continuing the last thread?

```py
mod.list("thread", desc="Which conversation thread to continue")
```

```talon
list: user.thread
-
and: continueLast
```

wolfmanstout:
It is a bit of both but more for current implementation. I could imagine that we add support for named threads, which might need more than just a list (although if it's just hardcoded conversation IDs I guess that could work). I generally try to avoid overdesigning for a future I haven't yet worked out, but this seemed useful enough in the near term just for the current purposes. If I put an optional literal in the command spec I don't have a way to figure out in the Talon file whether it was specified.

C-Loftus:
Makes sense, feel free to merge when you are happy

PR #178, opened 2025-04-04
Add models.json configuration for model-specific settings

wolfmanstout:
This change introduces:
- models.json configuration file support with auto-reload
- Model-specific API options and LLM options
- Model-specific system prompt capability
- Model ID mapping for alias support
- Deprecation of global model_temperature setting in favor of per-model config


C-Loftus:
Thanks for PR! I am going to test this out in more detail and then get you some comments tomorrow most likely.

lib/modelHelpers.py
```diff
@@ -18,6 +19,43 @@
...
+class ModelConfig(TypedDict):
+    name: str
+    model_id: NotRequired[str]
+    system_prompt: NotRequired[str]
+    llm_options: NotRequired[dict[str, Any]]
+    api_options: NotRequired[dict[str, Any]]
+
+
+# Path to the models.json file
+MODELS_PATH = Path(__file__).parent.parent / "models.json"
```
C-Loftus, line 32: Perhaps we could create a setting for this location to prevent potential merge conflicts for users / allow for easier git sync?

models.json.example
```diff
@@ -0,0 +1,25 @@
+// This is an example model configuration file.
+// To use: copy settings into models.json in the same directory. Remove any comments.
+[
+    {
+        // The name used in user.model_default or the right-hand-side of model.talon-list.
+        "name": "gpt-4o-mini",
+        // Additional JSON merged into the OpenAI Chat Completions API request if user.model_endpoint is not "llm".
+        "api_options": {
+            // The temperature of the model. Higher values make the model more creative.
+            "temperature": 0.7
```
C-Loftus, line 10: This documentation is all super great. Thanks! The only issue is that if the user copies this, the JSON parsing will fail since comments aren't valid in JSON like this. I wonder if it would make sense to have a key like `__META__` at the root or something analogous to allow for one large comment that just gets ignored in actual usage?

It is hacky but since json doesn't support comments I am not sure if there is a better way.

models.json.example
```diff
@@ -0,0 +1,25 @@
...
+        }
+    },
+    {
+        "name": "gemini-2.0-flash-search",
+        // The model ID used in the LLM CLI tool or the API. If unspecified, defaults to the name.
+        "model_id": "gemini-2.0-flash",
+        // Model-specific system prompt (overrides user.model_system_prompt).
+        "system_prompt": "You are a sassy but helpful assistant.",
+        // Options passed to the LLM CLI tool if user.model_endpoint = "llm". Run `llm models --options` to see all options.
```
C-Loftus, line 19:
```suggestion
        // Options passed to the LLM CLI tool if user.model_endpoint = "llm". Run `llm models --options` to see all options for each model
```

models.json.example
```diff
@@ -0,0 +1,25 @@
...
+        "llm_options": {
+            // Enables the Gemini search feature, which allows the model to search the web for information.
```
C-Loftus, line 21:
```suggestion
            // Enables a model-specific setting, namely, the Gemini search feature, which allows the model to search the web for information.
```

models.json
```diff
@@ -0,0 +1,11 @@
...
+    {
+        "name": "gpt-4o-mini",
+        "api_options": {
+            "temperature": 0.6
+        },
+        "llm_options": {
+            "temperature": 0.6
+        }
+    }
+]
```
C-Loftus, line 11: If we intend to have this as the default perhaps we should have this in the code and then just have the example file if they want to copy it? I say this just since a user might edit this and then end up having merge conflicts upon pulling.

C-Loftus (commented):
Thanks! This is super great; I tested myself locally with Llama 3. 

I had a few comments moreso on usability / UX. Not sure if there is a right / wrong answer for some of them so feel free to change / keep as you see fit

models.json.example
wolfmanstout: It's up to you, but I'll share my reasoning:
- I'm following the JSON with comments format that vscode created. So while Python will indeed not parse this, it is a somewhat standard format.
- I think that inline comments are significantly more readable than a comment at the top of the file.
- I do note that the user needs to remove comments (although I agree that this could easily be forgotten -- at least there would be a visible error in this case).

wolfmanstout:
Thanks for the review! 

The remaining comments here are all about models.json git merge issues. My assumption here was that people would simply maintain forks, but I understand that some people would prefer to keep customization separate. If you prefer, we could simply add this file to .gitignore, and generate a default version dynamically when the library is first loaded (it is important to have the file initialized so that Talon can watch for changes without requiring a restart). We could either use the defaults that I have defined here, or an empty config "[]". The only reason I used these defaults is that they preserve the existing default temperature. Otherwise I think the example file is sufficient.

If we do that, I don't think we need a setting to configure the location of the file, but let me know if you would still prefer to have that (in which case I would suggest using the default if empty and otherwise require an absolute path).

models.json.example
C-Loftus: Yeah I think you are right; comments probably are much cleaner. Let's keep it. 

Thinking out loud, my only worry was that less technical users might not be in an editor with syntax highlighting and be confused, but realistically this is probably more of a feature for advanced users so I am happy to keep it as is. 
I can add to docs more if needed.

C-Loftus:
Thank you James, much appreciated.

> My assumption here was that people would simply maintain forks, but I understand that some people would prefer to keep customization separate. 

Yeah I think if possible I would like this repo to not require any forking for config changes. @jaresty recommended this pattern and I feel it has worked out well. 

>  The only reason I used these defaults is that they preserve the existing default temperature. Otherwise I think the example file is sufficient.

Makes sense; in my opinion I don't need we need to worry about preserving that assuming the default temperature is reasonable (which I believe it is). Using the example file is good with me if that works for you!

> If we do that, I don't think we need a setting to configure the location of the file,

Think that seems fair to me.

wolfmanstout:
I think this PR is now ready to go. I merged in the Astro changes.

C-Loftus:
Looks good to me thanks!

PR #180, opened 2025-04-05
Import local instead of fetch in docs build; cleanup astro deps

C-Loftus:
- current astro docs fetch a remote URL; changed this to a local file so it is better for local dev and reviews
- cleanup old/insecure dependencies that could be fixed with `npm audit fix` (realistically this probably doesn't make a build deal since it is a static site)
- remove an old asset


.docs/src/components/LocalFile.astro
```diff
@@ -0,0 +1,13 @@
+---
+import { readFile } from "fs/promises";
+import { Code } from "@astrojs/starlight/components";
+const { data } = Astro.props; // Relative path like './src/data/example.yml'
```
wolfmanstout, line 4: Based on the usages, it looks like this is resolved relative to this file, so maybe change the example?

C-Loftus: Good catch; done

wolfmanstout (approved):

PR #185, opened 2025-06-17
Fix UTF encoding issues in send_request_to_llm_cli

wolfmanstout:
## Summary
- Fixes UTF encoding issues in the `send_request_to_llm_cli` function
- Configures process environment to enable UTF-8 mode on Windows
- Uses UTF-8 encoding consistently across all platforms
- Removes hardcoded cp1252 encoding that was causing character encoding problems

## Changes
- Added `PYTHONUTF8=1` environment variable for Windows Python 3.7+
- Changed encoding from conditional cp1252/utf-8 to always use utf-8
- Added environment parameter to subprocess.run call on Windows
- Updated comments to reflect the encoding fix

## Test plan
- [x] Code compiles without errors
- [x] Changes are focused only on UTF encoding fixes
- [x] No breaking changes to existing functionality
- [x] Windows users should see improved character encoding support

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/modelHelpers.py
+++ b/lib/modelHelpers.py
@@ -372,10 +372,14 @@ def send_request_to_llm_cli(
     if GPTState.debug_enabled:
         print(command)
 
+    # Configure output encoding
+    process_env = os.environ.copy()
+    if platform.system() == "Windows":
+        process_env["PYTHONUTF8"] = "1"  # For Python 3.7+ to enable UTF-8 mode
+    # On other platforms, UTF-8 is also the common/expected encoding.
+    output_encoding = "utf-8"
+
     # Execute command and capture output.
-    # Talon changes locale.getpreferredencoding(False) to "utf-8" on
-    # Windows, but the llm command responds with cp1252 encoding.
-    output_encoding = "cp1252" if platform.system() == "Windows" else "utf-8"
     try:
         result = subprocess.run(
             command,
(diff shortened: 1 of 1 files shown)


wolfmanstout:
My favorite way to test this is to request "generate poop emoji" 💩

lib/modelHelpers.py
```diff
@@ -372,10 +372,14 @@ def send_request_to_llm_cli(
     if GPTState.debug_enabled:
         print(command)
 
+    # Configure output encoding
+    process_env = os.environ.copy()
+    if platform.system() == "Windows":
```
C-Loftus, line 377: Seems reasonable to me! Feel free to merge when you are satisfied

Just curious, what is the purpose of copying the entire environment instead of just making a blank dict and setting just `PYTHONUTF8`? Are there other env vars we want to propagate?

wolfmanstout: The default behavior is to inherit the current environment, so this seemed like the more targeted change to existing behavior. I don't have any specific scenarios in mind where this is required although I suppose this gives the user some control through system environment variables like they would typically have. If you would prefer, I could try just setting the one entry to see if that works.

C-Loftus: Makes sense! Either way seems reasonable to me. If you want to change go for it otherwise this looks good to merge to me

PR #187, opened 2025-07-25
Add undo functionality for chat command

BoltonBailey:
This is a short PR to add a talonscript command to access the "Undo Request" action (for the chat interface, rather than the inline interface).

Please let me know if the command should be different or if I'm missing something else about a better way to do this.

### DIFF
--- a/copilot/copilot.talon
+++ b/copilot/copilot.talon
@@ -22,6 +22,7 @@ pilot bring <user.cursorless_ordinal_or_last> {user.makeshift_destination} <user
     user.cursorless_command(makeshift_destination, cursorless_target)
     user.copilot_bring_code_block(cursorless_ordinal_or_last)
 pilot chat [<user.prose>]$: user.copilot_chat(prose or "")
+pilot undo chat: user.vscode("workbench.action.chat.undoEdit")
 pilot {user.copilot_slash_command} <user.cursorless_target> [to <user.prose>]$:
     user.cursorless_command("setSelection", cursorless_target)
     user.copilot_inline_chat(copilot_slash_command or "", prose or "")


wolfmanstout:
I ended up largely rewriting the grammar in this file to something more consistent. Here are the most relevant lines:
https://github.com/wolfmanstout/talon-ai-tools/blob/feef2bcd8626a777f405a0e4689a04e9f7b89021/copilot/copilot.talon#L37-L39

In case it's helpful, here's the full diff of my changes (click on Files Changed and then hit enter again in the address bar to jump to the right file):
https://github.com/C-Loftus/talon-ai-tools/compare/main...wolfmanstout:talon-ai-tools:main#diff-4b276f6f3455cb692466372c960eba342baf7703b1222b3bb72752ad29b694cb

The copilot functionality keeps evolving so quickly that I've updated this grammar several times. In fact, some of these commands might be out of date since I've been using Claude Code for a while now. I'm not sure whether it still makes sense to separate "chat" and "edit" commands -- depends on whether VSCode still has distinct commands to do stuff like attach files to the chat.

C-Loftus:
Thanks both! I am happy to merge this if it would be helpful. I agree with James that the command grammar is evolving very quickly and so this file may also be worth rewriting. 

I am inclined to merge this if it is useful and then perhaps a future PR can be made if we want to make more significant edits to the grammar and command patterns themselves?

BoltonBailey:
Yes, I was playing around with it more and I noticed a few other actions it might be nice to have commands for. I am happy to go and make another PR to consolidate @wolfmanstout 's changes and any other relevant new actions I find.

PR #190, opened 2025-08-02
Resolve recently-introduced grammar conflict in copilot.talon.

wolfmanstout:
cc @BoltonBailey

I'm assuming that chatEditor.action.navigateNext will be used more than editor.action.inlineSuggest.showNext. The former is a much more recently-introduced command, however, so this might surprise some people.


BoltonBailey:
I agree that the chat one  seems like the one I would want to use more. I endorse this change.

BoltonBailey (approved):

C-Loftus (approved):

